PR #132, opened 2014-12-06
Feature/controllers 404

chrishoult:
As part of JOINDIN-519, @akrabat asked whether a Controller's handle() method should return false (as per existing behavior) or throw an Exception to represent a 404.

It is my view that ```false``` is a valid response from ApiController::handle (not that I can supply a proper use case) and so to use it to signal a 404 is potentially confusing. Exception throwing will be a lot more extensible, and the mechanism will also be the fastest way to stop processing a Request and move into the app's error handling logic.

Going forward, this will probably become a moot point when a Router is created with explicit routing rules (rather than using the ```handle()``` mechanism) as has been discussed with @lornajane.

### DIFF
--- a/src/controllers/ApiController.php
+++ b/src/controllers/ApiController.php
@@ -6,8 +6,11 @@ abstract class ApiController {
     /**
      * Handles a Request to this controller
      *
-     * @param Request $request
-     * @param mixed $db
+     * @param Request $request  The Request to respond to
+     * @param mixed $db         The PDO object
+     *
+     * @return mixed        The response to the handled Request. This is passed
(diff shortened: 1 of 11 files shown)


lornajane:
So should I just close this?

chrishoult:
I think that's best :)

On 3 January 2015 at 21:48, Lorna Jane Mitchell <notifications@github.com>
wrote:

> So should I just close this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/joindin/joindin-api/pull/132#issuecomment-68610906>.
>



-- 
Senior Software Engineer
M: 07812995301

PR #133, opened 2014-12-11
New regex-based router

chrishoult:
Contains regular expression-defined routes, explicitly named controllers and actions, and named parameters in URLs


lornajane:
THANKYOU for doing this :)  I am really looking forward to being able to get subresources separated out and doing all kinds of other things that just didn't fit with that very rigid routing that we had previously

PR #134, opened 2014-12-16
Fixes an issue with transfered dates

heiglandreas:
This Commit fixes an issue with dates that are first converted to
timestamps and these timestamps are then used to create a new DateTime
object. But as that expects timestamps to have a prepend '@' that has
been added.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -471,8 +471,8 @@ class EventsController extends ApiController {
                 try {
                     // make the timezone, and read in times with respect to that
                     $tz = new DateTimeZone($event['tz_continent'] . '/' . $event['tz_place']);
-                    $start_date = new DateTime('@' . strtotime($request->getParameter("start_date")), $tz);
-                    $end_date   = new DateTime('@' . strtotime($request->getParameter("end_date")), $tz);
+                    $start_date = new DateTime($request->getParameter("start_date"), $tz);
+                    $end_date   = new DateTime($request->getParameter("end_date"), $tz);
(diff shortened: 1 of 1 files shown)


src/controllers/EventsController.php
```diff
@@ -499,12 +499,12 @@ public function putAction($request, $db)
             }
             $cfp_start_date = $request->getParameter("cfp_start_date", false);
             if (false !== $cfp_start_date && strtotime($cfp_start_date)) {
-                $cfp_start_date = new DateTime(strtotime($cfp_start_date), $tz);
+                $cfp_start_date = new DateTime('@' . strtotime($cfp_start_date), $tz);
```
akrabat: I'm nearly sure that the correct solution is to remove the `strtotime` as `DateTime`'s constructor has that functionality built in.

heiglandreas: You are definitely right there. As far as I recall my intention was to keep the the original code as untouched as possible. I was wondering why that was done in that way but thought it had it's reason. Therefore I just kept the strange code and worked around it.

But it seems that the ```strtotome``` wasn't necessary in the first place. Therefore I'll remove it.

lornajane: Never assume that we knew what we were doing at any point in the past :)

heiglandreas: Yeah, I got that one :wink:

heiglandreas:
This seemst to work out now.

heiglandreas:
Thanks!

PR #137, opened 2015-01-09
Indicate if this user can edit this event

lornajane:


heiglandreas:
@lornajane That's looking great! Makes it much easier to determine whether one can edit the event.

PR #142, opened 2015-01-10
Add icon_url into event output

richsage:
This adds the full URL to the event's icon filename, if there is one (field is blank otherwise) - saves any clients having to know what to prefix to the icon filename.This adds the full URL to the event's icon filename, if there is one (field is blank otherwise) - saves any clients having to know what to prefix to the icon filename themselves, and allows moving of assets at a later date if required.

### DIFF
--- a/src/models/EventMapper.php
+++ b/src/models/EventMapper.php
@@ -32,6 +32,7 @@ class EventMapper extends ApiMapper
             'tracks_count' => 'track_count',
             'talks_count' => 'talk_count',
             'icon' => 'event_icon',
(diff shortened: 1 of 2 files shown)


lornajane:
We need to think about this - the paths will change or might change when we ditch web1, those icons need to come over to web2 at some point.  I resisted putting in the URLs until we were sure we knew what we were doing.  We should probably write down what the plan is for ending up with URLs on the API that point to web2 after its moved to being the main/only site

richsage:
OK, makes sense. I'll close this for now then.

PR #144, opened 2015-01-19
Enable dispatching to TalksController::postAction

akrabat:
Now that dispatching is an external operation, postAction needs to be public.

Fixes JOINDIN-536.


lornajane:
Thanks! I also added a test for the tracks endpoint

PR #148, opened 2015-02-02
Allow the generated test users to login

magicmonkey:


tools/dbgen/generator.class.php
```diff
@@ -588,7 +588,7 @@ protected function _generateUsers($count) {
 
         // Ima admin
         $password = '$2y$10$StMpCySaSLHuvh3KLFNiZ.vSEykBF8kVWEG6ZtuXzQd4F9WWlfBFi';
-        echo "INSERT INTO `user` (`username`, `password`, `email`, `last_login`, `ID`, `verified`, `admin`, `full_name`, `active`, `twitter_username`, `request_code`) VALUES\n";
+        echo "INSERT INTO `user` (`username`, `password`, `verified`, `email`, `last_login`, `ID`, `verified`, `admin`, `full_name`, `active`, `twitter_username`, `request_code`) VALUES\n";
```
akrabat: I'm confused as `verified` is already in the list (4 columns further down).

tools/dbgen/generator.class.php
```diff
@@ -608,8 +608,8 @@ protected function _generateUsers($count) {
             $user->twitter = $this->_chance(USER_HAS_TWITTER) ? "@".$user->username : "";
 
             echo ",\n";
-            printf ("('%s', '%s', '%s', %d, %d, %d, %d, '%s', %d, '%s', NULL)",
-                                   $user->username, $user->password, $user->email, $user->last_login, $user->id, $user->verified, $user->admin, $user->fullname, $user->active, $user->twitter);
+            printf ("('%s', '%s', '%s', '%s', %d, %d, %d, %d, '%s', %d, '%s', NULL)",
+                                   $user->username, $user->password, 1, $user->email, $user->last_login, $user->id, $user->verified, $user->admin, $user->fullname, $user->active, $user->twitter);
```
akrabat: Ditto here:  `$user->verified` is being set, I think?

tools/dbgen/generator.class.php
magicmonkey: Yep you're right, it was only the seed.sql file which was causing me problems.  It was unfair of me to assume it would be consistently broken :-/  I'll change this PR to only be for the seed.sql file.

tools/dbgen/generator.class.php
magicmonkey: Ditto above.

magicmonkey:
PR updated to only affect seed.sql.

PR #151, opened 2015-02-21
Addition to generate friendly URL tokens for events

phoenixrises:
At the moment when generating random event data for dev we don't populate the friendly URL column. This change generates the inflected name and generates a random byte -> hex string to append to the string then checks against a tracking variable to make sure it's unique.


akrabat:
ok to test

phoenixrises:
This change is simply for development, it only affects use of dbgen, it's so that when you generate the data to populate the DB, you get the URLs too. You're not guaranteed to go through the process that will generate them through the API if that's not what's being worked on, if that makes sense.

As a side note, is there a reason for putting off the generation until an API call is made rather than on event creation?

lornajane:
The API is the only way to access these fields, the logic is there because we had a lot of existing talks and we weren't ready to move talk creation over to the api/web2 yet so we just populate the field the first time it comes through the API, and I still think that's a good solution so I'm closing the PR.

PR #155, opened 2015-02-22
Some DB indexes

magicmonkey:
From observing the slow query log, there are numerous queries which would benefit from an index.  This will help some of the more obvious ones I spotted.


akrabat:
Note that patch53.sql is now in master (due to #149), so update this to 54 when merging.

PR #156, opened 2015-02-26
Joindin 544

ThijsFeryn:
I took @akrabat's comments into account, fixed the Frisby indentation and put my DB patches in patch55.sql.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -214,6 +214,7 @@ class EventsController extends ApiController {
                             400
                         );
                     }
+                                        
                     // no anonymous comments over the API
                     if(!isset($request->user_id) || empty($request->user_id)) {
                         throw new Exception('You must log in to comment');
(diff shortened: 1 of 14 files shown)


akrabat:
ok to test

src/controllers/EventsController.php
```diff
@@ -222,7 +223,12 @@ public function postAction($request, $db) {
                     $users = $user_mapper->getUserById($request->user_id);
                     $thisUser = $users['users'][0];
 
-                    $commentText = filter_var($request->getParameter('comment'), FILTER_SANITIZE_STRING);
```
lornajane: Why remove the sanitize filter on this variable?

lornajane:
I'm really happy with how this pull request is coming together, thanks Thijs!  I've added one tiny question and there is one thing I would like to change.

I'd prefer the ratings average to be calculated in SQL in the API, rather than in a stored procedure on the database.  The stored procedures are harder to maintain, and while I realise this PR just follows the pattern of the talk averages, this feature won't be backported to web1 so I'd prefer to just have the SQL straight into the API, please.

Apart from that, I think we're pretty much there at least with the API component, I am very happy to see the additions to the sample data generator and the tests as well, great job!!

src/controllers/EventsController.php
ThijsFeryn: Quotes weren't displayed correctly.  I checked and talk comments don't use this filter. Either we remove the filter for comments or we make sure the filtered version is displayed correctly on screen.

ThijsFeryn:
Hi @lornajane 
I refactored it a bit and made sure the stored proc is gone. I didn't like to subqueries in the SQL statement, so I turned it into left joins. 

Hope you don't mind.

src/controllers/EventsController.php
ThijsFeryn: Maybe we should consider this a web2 issue rather than an api issue. Should I restore the filter and open an issue for web2?

lornajane:
Woooo!  Thanks Thijs :)

PR #157, opened 2015-02-28
Jira 482 password change

mgdm:
An update of @cyberspice's branches to pull in master. Verified to work.


src/config/routes/2.1.json
```diff
@@ -84,6 +84,14 @@
         "action": "handle",
         "verbs": ["GET"]
     },
+
+    {
+        "path": "/password/?$",
+        "controller": "PasswordController",
+        "action": "handle",
```
akrabat: Don't use the `handle` action - use a specific method as has been done for [usernameReminder](https://github.com/joindin/joindin-api/blob/master/src/config/routes/2.1.json#L95-L100)

src/controllers/PasswordController.php
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @return mixed
+     */
+    public function postAction(Request $request, $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to change your password", 400);
+        }
+
+        $userId    = $request->getUserId();
```
akrabat: We should probably have a policy on whether we're using `$request->user_id` or `$request->getUserId()`. We appear to use the property directly everywhere else. Discussion required.

akrabat:
We shouldn't be asking the client to send through its client_id and secret for this call as it is not part of OAuth2 authentication process.

Instead, given that we have the user's token, we should look up the `consumer_key` in `oauth_access_tokens' and from there check `oauth_consumers` to see if this client is allowed to change the password. I suspect we'll either need a new column on `oauth_consumers`or we could use `enable_password_grant` if we're happy with all trusted clients being able to change the password.

Thoughts?

lornajane:
I agree that this should be done for a logged in user and so use their token.  I hadn't really thought about limiting which clients have the ability to do this, only clients to whom it makes sense would implement it and I can't see a security problem since they have to be logged in?  You could argue that it makes no sense so only the privileged consumers can do it - in which case let's say that if you can log in to a consumer with the password grant then you can also change the password for the user.

src/controllers/PasswordController.php
```diff
@@ -0,0 +1,80 @@
...
+        $password = $request->getParameter('password');
+	if (empty($password)) {
+            throw new Exception('The field "password" is required and must not be empty', 400);
+        }
+
+        $clientId     = $request->getParameter('client_id');
+        $clientSecret = $request->getParameter('client_secret');
+        if (!$this->oauthModel->isClientPermittedPasswordGrant($clientId, $clientSecret)) {
+            throw new Exception("This client cannot authenticate using the password grant type", 403);
+        }
```
akrabat: Remove from `$grantType = $request->getParameter('grant_type');` to here.

Instead, use the Request's `getAccessToken()` method to give you a valid value for the `access_token` column in the `auth_access_tokens` table which will allow you to join through to `auth_consumers` table via the `consumer_key` column so that you can get the value of the `enable_password_grant` column. If it's not 1, then allow a password change, otherwise do not allow it.

src/controllers/PasswordController.php
```diff
@@ -0,0 +1,80 @@
...
+
+        if (!$this->oauthModel->setPasswordForUserId($userId, $password)) {
+            throw new Exception("Update of password failed", 403);
+        }
+
+        if (isset($this->config['oauth']['expirable_client_ids'])) {
+            $this->oauthModel->expireOldTokens($this->config['oauth']['expirable_client_ids']);
+        }
```
akrabat: Don't need to expireOldTokens when changing a password, so lose this `if` statement.

src/controllers/PasswordController.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $accessToken = $this->oauthModel->newAccessToken($clientId, $userId);
+        if ($accessToken) {
+             return array("access_token" => $accessToken);
+        }
+
+        throw new Exception("Signin failed", 403);
```
akrabat: This lot can go too as we're not issuing new access token for this trusted client.

lornajane:
I'm going to pick this up as I don't think mgdm has time ... will close and re-open when I've picked up on all the comments here

lornajane:
Replaced by #168 which is also rebased

PR #158, opened 2015-03-08
Add the ability for tests to create an automatically verified user if the config permits this

lornajane:


src/controllers/UsersController.php
```diff
@@ -148,6 +148,16 @@ public function postAction($request, $db){
                 $user_id = $user_mapper->createUser($user);
                 header("Location: " . $request->base . $request->path_info . '/' . $user_id, NULL, 201);
 
+                // autoverify for test platforms
+                if(isset($this->config['features']['allow_auto_verify_users']) 
+                    && $this->config['features']['allow_auto_verify_users']) {
+                        if($request->getParameter("auto_verify_user") == "true") {
```
akrabat: Why a separate if statement rather than an && on the current one.

Also, should auto_verify_user be set to "yes" for consistency with verbose?

src/models/UserMapper.php
```diff
@@ -371,4 +371,21 @@ public function getUserIdFromEmail($email) {
         }
         return false;
     }
+
+    /**
+     * Used only on test platforms, if the config is enabled
+     * 
+     * Designed to allow creation of verified users for testing purposes
+     */
+    public function verifyThisTestUser($user_id) {
```
akrabat: Ideally, I think that this method should be called verifyThisUser($user_id) and then used within verifyUser(), so that we don't duplicate this code.

If we want to have the word "test" in the UserMapper's public api, then verifyThisUser() could be protected/private and verifyThisTestUser() could just call it,

src/controllers/UsersController.php
lornajane: separate statements because I like feature toggles to be separate.  Which is another way of saying: no reason, that's just how my brain works.

PR #165, opened 2015-03-14
Useful administrative flags on the user record

akrabat:
Add `can_edit` to the user record if the logged in user is a site admin or this is the current user's record.
Also expose the `admin` field for every user to site admins only.
Both fields are only available if a user is logged in.

### DIFF
--- a/src/models/UserMapper.php
+++ b/src/models/UserMapper.php
@@ -164,8 +164,19 @@ class UserMapper extends ApiMapper
             }
 
             foreach ($results as $key => $row) {
+                // can the logged in user edit this user?
+                $canEdit = false;
+                if ($userIsSiteAdmin || $row['ID'] == $this->_request->user_id) {
+                    $canEdit = true;
+                }
+                
                 if (true === $verbose) {
                     $list[$key]['gravatar_hash'] = md5(strtolower($row['email']));
+
+                    // if this record can be edited, then expose the email address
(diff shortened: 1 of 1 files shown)


src/models/UserMapper.php
```diff
@@ -174,6 +196,10 @@ public function transformResults($results, $verbose)
                     . $row['ID'] . '/hosted/';
                 $list[$key]['talk_comments_uri'] = $base . '/' . $version . '/users/'
                     . $row['ID'] . '/talk_comments/';
+                
+                if ($verbose && isset($this->_request->user_id)) {
```
heiglandreas: Can it be possible that ```$this->_request->user_id``` is **not** set at this point? If so, what happens in line 169?

Or could the ```can_edit```-flag be set there as well?

lornajane: If you're not logged in, then this field won't be set.  This function could be used by both logged in and anonymous users I think, but only if you are logged in and have rights would we want to add this field.

heiglandreas: Ok. Got that. But shouldn't the ```isset``` then be added to line 169 as well?

akrabat: Almost certainly.

akrabat: $user_id is a public property of Request, so it always exists. It's either null or a number.

PR #168, opened 2015-03-20
Allow trusted consumers to change user's password via the API

lornajane:
This picks up from the work done by cyberspice/mgdm


lornajane:
Updated AGAIN to include editing other user account fields, not just the password, so this is now replaced by https://github.com/joindin/joindin-api/pull/169

PR #169, opened 2015-03-24
Users can edit their own accounts

lornajane:
Allow the user or a site admin to edit the fields on a user record: full name, email, twitter username and password.


src/models/UserMapper.php
```diff
@@ -385,4 +385,80 @@ protected function markUserVerified($user_id) {
...
+            'full_name',
+            'email',
+        );
+        $contains_mandatory_fields = !array_diff($mandatory_fields, array_keys($user));
+        if (!$contains_mandatory_fields) {
+            throw new Exception("Missing mandatory fields");
+        }
+
+        // encode the password
+        $user['password'] = password_hash(md5($user['password']), PASSWORD_DEFAULT);
```
akrabat: password is not mandatory and may not have been sent in, so please check that it exists before using.

src/models/UserMapper.php
```diff
@@ -385,4 +385,80 @@ protected function markUserVerified($user_id) {
...
+
+        $sql = "update user set ";
+
+        // create list of column to API field name for all valid fields
+        $fields = $this->getVerboseFields();
+        // add the fields that we don't expose for users
+        $fields["email"] = "email";
+        $fields["password"] = "password";
```
akrabat: Same as above. Password is not mandatory.

heiglandreas:
just a stupid question: Why are we pre-hasing the password to md5?

lornajane:
No such thing as a stupid question!  It's because the v1 API uses md5s so when we re-hashed the passwords, we couldn't store them un-md5-ed

heiglandreas:
So the passwords currently are md5ed and then hashed. which would mean we'd have to know whether the hash is from an md5ed password or directly from a password to verify a users login. And that would only be possible when we somewhere log whether the password has been "un-md5ed" (ugly word) before hashing so we know which way to verify. 

Would be an interesting task. Adding a field ```password_is_md5ed``` to the users-table which is by default 0 and after "un-md5ing" would be set to 1. And on login the password is checked either md5ed (field is 0) or un-md5ed (field is 1). And on successfull login the password is rehashed without md5 and the field is set to 1. Not that important though as it currently works. But it would mean we could also know who didn't log in since the change as the ```password_is_md5ed```-field would be still 0

Shall I add a task to jira?

akrabat:
The correct solution is to turn off the v1 API which we will do as soon as the iPhone app is updated to use v2.1.

src/controllers/UsersController.php
```diff
@@ -168,4 +168,107 @@ public function postAction($request, $db){
...
+                    }
+                }
+            }
+
+            // Optional Fields
+            $twitter_username = $request->getParameter("twitter_username", false);
+            if(false !== $twitter_username) {
+                $twitter_username = filter_var(trim($twitter_username), FILTER_SANITIZE_STRING);
+                if($twitter_username) {
+                    $user['twitter_username'] = $twitter_username;
```
akrabat: We don't need to set the default to `false`, test that and then filter and test again. Instead, don't set the default at all (as then it's an empty string), so we can skip the `false` test and just filter_var it directly.

src/models/UserMapper.php
```diff
@@ -385,4 +385,80 @@ protected function markUserVerified($user_id) {
...
+     *
+     * User must be the edited user, or a site admin
+     *
+     * @param int $user_id The identifier for the user to edit
+     * @return bool True if the user has privileges, false otherwise
+     */
+    public function thisUserHasAdminOn($user_id) {
+        // do we even have an authenticated user?
+        $loggedInUser = $this->_request->getUserId();
+        if(false !== $loggedInUser) {
```
akrabat: `$this->_request->getUserId()` will return null or a number. I don't think it will ever return a false?
You can remove the "if (false"  test anyway as the normal path through this method is a valid logged in user.

akrabat:
Merged in [commit 3200052](https://github.com/joindin/joindin-api/commit/3200052ed6e0b28fe3daebcca4883bec7b98edc0).

PR #173, opened 2015-04-03
replaced deprecated --strict 

damko:


lornajane:
ok to test

lornajane:
Is there a range of versions of PHPUnit where this is the right fix to use?  I am not certain what version either the development VMs or the build server uses but I doubt it would be very new in either case

akrabat:
The server has PHPUnit 4.1.3.
The VM has PHPUnit 4.5.0.

damko:
The fix is good for >= 4.5.0 but it doesn't break on phpunit 4.0.7 (I checked). Of course, versions < 4.5.0 will run without the --strict option

lornajane:
Looks good but can I have just the config changes on their own, please?  I don't want to merge a work-in-progress unit test.

lornajane:
I merged the phpunit change, but not the work in progress test.  For your next commit, try this:
 - create a new branch
 - make the changes for one feature and submit it
 - start a new branch for the next idea
 - don't make changes to your master branch, they'll clash with ours (ping us in channel if you want help cleaning up your master branch at this point)

Thanks for contributing!!

PR #183, opened 2015-04-19
Prevent hosts from rating their own events

akrabat:
If a host submits a comment that has a rating, then ignore that rating.


lornajane:
I'm struggling to see this working. I created a new user, registering via web2, and logged in and submitted a new event also via web2.  Approved it on web1 and found that I can comment and rate, and the rating gets stored, when I comment via web2.

akrabat:
That's strange. I've had a play this morning and I cannot reproduce.

PR #185, opened 2015-04-26
Add API endpoints for forgot password

lornajane:
This adds two endpoints:
 - one to request a password reset email be sent to the user
 - one to accept the token sent in that email, and accept a new password for this user (it also deletes the token)

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -267,4 +267,26 @@ class UsersController extends ApiController {
         }
         throw new Exception("Could not update user", 400);
     }
+
+    public function passwordReset(Request $request, $db)
+    {
+        $token = filter_var($request->getParameter("token"), FILTER_SANITIZE_STRING);
+        if(empty($token)) {
+            throw new Exception("Reset token must be supplied", 400);
+        }
+
+        $password = $request->getParameter("password");
+        if(empty($password)) {
+            throw new Exception("New password must be supplied", 400);
+        }
+
+        $user_mapper = new UserMapper($db, $request);
+        $success = $user_mapper->resetPassword($token, $password);
+        if($success) {
+            header("Content-Length: 0", NULL, 204);
+            exit; // no more content
+        } else {
+            throw new Exception("Verification failed", 400);
+        }
+    }
 }
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -279,14 +279,22 @@ class UsersController extends ApiController {
         if(empty($password)) {
             throw new Exception("New password must be supplied", 400);
         }
-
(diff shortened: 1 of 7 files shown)


src/models/UserMapper.php
```diff
@@ -488,5 +488,116 @@ public function editUser($user, $userId) {
...
     }
 
+    /**
+     * Function to get just the user ID
+     *
+     * @param string $username The username of the user we're looking for
+     * @return $user_id The user's ID (or false, if we didn't find her)
+     */
+    public function getUserIdFromUsername($username) {
+        $sql = "select ID from user where username = :username";
```
akrabat: `getUserByUsername()` also checks for active=1. This method should do the same I think.

src/controllers/UsersController.php
```diff
@@ -267,4 +267,26 @@ public function updateUser(Request $request, $db)
...
+        $token = filter_var($request->getParameter("token"), FILTER_SANITIZE_STRING);
+        if(empty($token)) {
+            throw new Exception("Reset token must be supplied", 400);
+        }
+
+        $password = $request->getParameter("password");
+        if(empty($password)) {
+            throw new Exception("New password must be supplied", 400);
+        }
+
```
akrabat: Need to call `checkPasswordValidity()` here as per UsersController::postAction().

src/models/UserMapper.php
```diff
@@ -488,5 +488,116 @@ public function editUser($user, $userId) {
...
+     * Check that the token is valid, find out which user this is, save their
+     * new password and then delete their other tokens
+     *
+     * @param string $token    The reset we sent them by email (link goes to web2)
+     * @param string $password The new password they chose
+     */
+    public function resetPassword($token, $password) {
+        // does the token exist, and whose is it?
+        $select_sql = "select user_id from password_reset_tokens "
+            . "where token = :token";
```
akrabat: We should have a time limit in here against the `created_date` column. A token that's a year old certainly shouldn't work.

src/models/UserMapper.php
```diff
@@ -488,5 +488,116 @@ public function editUser($user, $userId) {
...
+                $update_response = $update_stmt->execute($update_data);
+
+                if($update_response) {
+                    // delete all the user's tokens; they don't need them now
+                    $delete_sql = "delete from password_reset_tokens "
+                        . "where user_id = :user_id";
+
+                    $stmt = $this->_db->prepare($delete_sql);
+                    $stmt->execute(array("user_id"  => $user_id));
+
```
akrabat: Arguably, we should delete all really old tokens, but I don't think it's an impediment to the PR as it'll only worry us if the web2 form or endpoint gets spammed. Possibly create an issue if not addressed now.

PR #186, opened 2015-05-01
Create and Edit talks via the API

heiglandreas:
This PR shall allow to create, edit and delete talks via the API as described in JOINDIN-525.

Currently the following is possible:
* Create a talk without adapting the already available ```TalkMapper::save()```- Method
* Had to adapt the ```TalkMapper::save()``` to allow all available fields to be saved.
* Edit a talk using a ```PUT```-request.
* Delete a talk using a ```DELETE```-request

The expected field-names are the same as the ones that are output via the get-method. 

An (Event)-Admin can edit all available fields whereas an approved speaker can only edit the title, the description, the language and the slides-link.

Still needed: Tests (unittests and frisby)!


heiglandreas:
The content of the PR is functional, tests are still missing, they will be the next thing on the agenda! Any feedback is very welcome!

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+            if (empty($talk['event_id'])) {
+                $errors[] = '"event_id" is a required field';
+            }
+
+            $talk['talk_title'] = filter_var($request->getParameter("talk_title"), FILTER_SANITIZE_STRING);
+            if(empty($talk['talk_title'])) {
+                $errors[] = "'talk_title' is a required field";
+            }
+
+            $talk['url_friendly_talk_title'] = filter_var($request->getParameter("ur_friendly_talk_title"), FILTER_SANITIZE_STRING);
```
akrabat: url_friendly_talk_title is generated by api, not supplied by the user.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+            $talk['language'] = filter_var($request->getParameter('language'), FILTER_SANITIZE_STRING);
+            if (empty($talk['language'])) {
+                $talk['language'] = 'English - UK';
+            }
+            // TODO: @heiglandreas Check whether the language actually exists in the database
+            // When the language doesn't exist, the talk will not be found
+
+            $talk['slides_link'] = filter_var($request->getParameter('slides_link'), FILTER_SANITIZE_URL);
+
+            $incoming_speakers_list = $request->getParameter('speakers');
```
akrabat: I would have expected speakers to be a sub resource of the talk, though I'll defer to @lornajane's thoughts on this.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+                }, $incoming_speakers_list);
+            }
+
+            $talk_mapper = new TalkMapper($db, $request);
+            $new_id = $talk_mapper->save($talk);
+
+            $uri = $request->base . '/' . $request->version . '/talks/' . $new_id;
+            header("Location: " . $uri, true, 201);
+            exit;
+        //    throw new Exception("method not yet supported - sorry");
```
akrabat: Comment can be removed.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+     * * int event_id
+     * * string title
+     * * string url_friendly_talk_title
+     * * string description
+     * * string language (has to be a value from the langauges-table
+     * * string slides_link (has to be a URL)
+     * * string type (has to be a value from the categories-table
+     * * int duration
+     * * string date (will be parsed as date)
+     * * string[] speakers (speaker names as the fullnames in the user-table)
```
akrabat: Use hyphen for bullet points within docblocks as per https://github.com/joindin/joindin-api/blob/master/src/routers/VersionedRouter.php#L17-L22

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+     * @param $request
+     * @param $db
+     *
+     * @throws Exception
+     */
+    public function putAction($request, $db)
+    {
+        $errors = array();
+        if(!isset($request->user_id)) {
+            throw new Exception("You must be logged in to edit data", 400);
```
akrabat: Should this be a 403?

src/controllers/TalksController.php
```diff
@@ -2,16 +2,18 @@
 
 class TalksController extends ApiController {
     public function handle(Request $request, $db) {
```
akrabat: We aren't using the `handle` method any more as we route directly in `2.1.json`, so no need to change it. At some point, we should remove the abstract method from `ApiController`…

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+        }
+    }
+
+    /**
+     * Edit a talk
+     *
+     * This action expects the following parameters:
+     *
+     * * int talk_id
+     * * int event_id
```
akrabat: `talk_id` and `event_id` are not parameters when editing a talk. This is because we are PUTing to the talk resource directly, so the URI already has enough information in it for us to identify the talk and event ids.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+     * * string title
+     * * string url_friendly_talk_title
+     * * string description
+     * * string language (has to be a value from the langauges-table
+     * * string slides_link (has to be a URL)
+     * * string type (has to be a value from the categories-table
+     * * int duration
+     * * string date (will be parsed as date)
```
akrabat: Note in this bullet point that that we expect the `date` to include the time as well as the date.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+    public function putAction($request, $db)
+    {
+        $errors = array();
+        if(!isset($request->user_id)) {
+            throw new Exception("You must be logged in to edit data", 400);
+        }
+        $talk_id = $this->getItemId($request);
+
+        if (isset($request->url_elements[4])) {
+            // Not editing the talk but something "below"
```
akrabat: You should be able to craft the `path` in `2.1.json` so that we don't need this test at all. i.e. let's ensure that the router doesn't match sub-resources to this action method.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+            throw new Exception('Operation not supported, sorry', 404);
+        }
+
+        $talk_mapper = new TalkMapper($db, $request);
+
+        $existing_talk = $talk_mapper->getTalkById($talk_id, true);
+        if (! $existing_talk || $existing_talk['meta']['count'] < 1) {
+            throw new Exception(sprintf(
+                'There is no talk with ID "%s"',
```
akrabat: Really, we shouldn't be encouraging decoding of URLs, so change this message to "Talk not found." or similar.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+        $talk['slides_link'] = filter_var($request->getParameter('slides_link'), FILTER_SANITIZE_URL);
+
+        if (! $isAdmin) {
+            $talk['url_friendly_talk_title'] = $existing_talk['url_friendly_talk_title'];
+            $talk['type']               = $existing_talk['type'];
+            $talk['duration']           = $existing_talk['duration'];
+            $talk['start_date']               = (new \DateTime($existing_talk['start_date']))->format('U');
+        } else {
+
+            $talk['url_friendly_talk_title'] = filter_var(
```
akrabat: See above.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+            $talk['talk_description']  = filter_var($request->getParameter("talk_description"), FILTER_SANITIZE_STRING);
+            if (empty($talk['talk_description'])) {
+                $errors[] = "'talk_description' is a required field";
+            }
+
+            $talk['type']  = filter_var($request->getParameter("type"), FILTER_SANITIZE_STRING);
+            if (empty($talk['type'])) {
+                $errors[] = "'type' is a required field";
+            }
+            // TODO @heiglandreas check whether the type already exists in the database.
```
akrabat: Implement from `putAction` below…

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+            );
+            if (empty($talk['duration'])) {
+                $talk['duration'] = 60;
+            }
+
+            $talk['language'] = filter_var($request->getParameter('language'), FILTER_SANITIZE_STRING);
+            if (empty($talk['language'])) {
+                $talk['language'] = 'English - UK';
+            }
+            // TODO: @heiglandreas Check whether the language actually exists in the database
```
akrabat: Implement as per `putAction` below.

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
+        if (! in_array($talk['language'], $talk_mapper->getLanguages())) {
+            $errors[] = sprintf('The language "%s" isn\'t known', $talk['language']);
+        }
+        $talk['slides_link'] = filter_var($request->getParameter('slides_link'), FILTER_SANITIZE_URL);
+
+        if (! $isAdmin) {
+            $talk['url_friendly_talk_title'] = $existing_talk['url_friendly_talk_title'];
+            $talk['type']               = $existing_talk['type'];
+            $talk['duration']           = $existing_talk['duration'];
+            $talk['start_date']               = (new \DateTime($existing_talk['start_date']))->format('U');
```
akrabat: Either align the equals signs or don't :)

src/controllers/TalksController.php
```diff
@@ -140,8 +142,229 @@ public function postAction($request, $db) {
...
         }
+
+        if ($errors) {
+            throw new Exception(implode(". ", $errors), 400);
+        }
+
+        $new_id = $talk_mapper->edit($talk, $talk_id);
+
+        $uri = $request->base . '/' . $request->version . '/talks/' . $new_id;
+        header("Location: " . $uri, true, 201);
```
akrabat: We haven't created the resource, so rather than a `201`, how about a `204` as per [RFC2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6) ?

src/models/TalkMapper.php
```diff
@@ -313,25 +313,52 @@ public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose =
...
+     * * date (a timestamp)
+     * * duration
+     * * speakers (an array of names)
+     * * category (a value from the column categories:title)
+     *
+     * @param $data
+     *
+     * @return string
+     */
     public function save($data) {
```
akrabat: Let's call this method `createTalk` to match the EventMapper.

src/models/TalkMapper.php
```diff
@@ -313,25 +313,52 @@ public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose =
...
+     * * slides_link
+     * * language (a value from the column lang:lang_name
+     * * date (a timestamp)
+     * * duration
+     * * speakers (an array of names)
+     * * category (a value from the column categories:title)
+     *
+     * @param $data
+     *
+     * @return string
```
akrabat: We're returning the `talk_id` which is an integer.

src/controllers/TalksController.php
lornajane: Why is there a new ID?  The 204 without a location header is all we need here I think.

akrabat:
Note that we can already create a talk here: https://github.com/joindin/joindin-api/blob/master/src/controllers/EventsController.php#L133-L207

akrabat:
Please put the create new talk action into a new method called `TalksController::create` and add a new route that responds to POST of `/events/([0-9])+/talks` that dispatches to it.

src/models/TalkMapper.php
```diff
@@ -313,25 +313,52 @@ public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose =
...
+            . ':slides_link, (select ID from lang where lang_name = :language), '
+            . ':date, :duration)';
 
         $stmt = $this->_db->prepare($sql);
         $response = $stmt->execute(array(
             ':event_id' => $data['event_id'],
-            ':talk_title' => $data['title'],
-            ':talk_description' => $data['description'],
+            ':talk_title' => $data['talk_title'],
+            ':url_friendly_talk_title' => $data['url_friendly_talk_title'],
```
akrabat: Don't need url_friendly_talk_title here.

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+            }
+        }
+
+        $items['talk_id'] = $talk_id;
+
+        $stmt = $this->_db->prepare(sprintf($sql, implode(', ', $pairs)));
+
+        if (! $stmt->execute($items)) {
+            throw new Exception(sprintf(
+                'executing "%s" resulted in an error: %s',
```
akrabat: We should make sure that the SQL we're trying to execute doesn't leak through to the client. If this exception is passed directly to the client, then change it.

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+     * * category (a value from the column categories:title)
+     *
+     * @param array $talk    talk data to insert into the database.
+     * @param int   $talk_id The ID of the talk to be edited
+     *
+     * @return integer|false
+     */
+    public function edit($talk, $talk_id)
+    {
+        var_dump($talk);
```
akrabat: This can be deleted.

src/models/TalkMapper.php
```diff
@@ -313,25 +313,52 @@ public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose =
...
         ));
         $talk_id = $this->_db->lastInsertId();
 
+        if (0 == $talk_id) {
+            throw new Exception(sprintf('There has been an error storing the talk.'), 400);
+        }
+
         // set talk type
         // TODO support more than just talks
         $cat_sql = 'insert into talk_cat (talk_id, cat_id) values (:talk_id, 1)';
```
akrabat: This will need to be completed before merge.

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+        if (! $stmt->execute($items)) {
+            throw new Exception(sprintf(
+                'executing "%s" resulted in an error: %s',
+                $stmt->queryString,
+                implode(' :: ', $stmt->errorInfo())
+            ));
+            return false;
+        }
+
+        if (isset($talk['category'])) {
```
akrabat: Should this be testing against `$data`? I should be able to add a category to talk that doesn't have any.

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+     *
+     * * talk_title
+     * * url_fiendly_talk_title
+     * * talk_description
+     * * slides_link
+     * * language (a value from the column lang:lang_name
+     * * start_date (a timestamp)
+     * * duration
+     * * speakers (an array of names)
+     * * category (a value from the column categories:title)
```
akrabat: Isn't this `type`, not `category` ? (I'm looking at https://github.com/joindin/joindin-api/pull/186/files#diff-e27b217374c8436f4b830c74ed46cbbaR323)

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+            return false;
+        }
+
+        if (isset($talk['category'])) {
+            $cat_sql = 'select id from talk_cat where talk_id = :talk_id and cat_id = :cat_id';
+            $cat_stmt = $this->_db->prepare($cat_sql);
+            $result = $cat_stmt->execute(array(
+                ':talk_id' => $talk_id,
+                ':cat_id'  => $talk['type'],
+            ));
```
akrabat: There's something wrong here. `$talk['type']` is the text of the category, but cat_id is an integer…

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+                $cat_stmt = $this->_db->prepare($cat_sql);
+                $cat_stmt->execute(array(
+                    ':talk_id' => $talk_id,
+                    ':cat_id'  => $talk['type'],
+                ));
+                $cat_sql = 'insert into talk_cat (talk_id, cat_id) values (:talk_id, :cat_id)';
+                $cat_stmt = $this->_db->prepare($cat_sql);
+                $cat_stmt->execute(array(
+                    ':talk_id' => $talk_id,
+                    ':cat_id'  => $talk['type'],
```
akrabat: Does this work? `$talk['type]'` is a string I believe…

src/models/TalkMapper.php
```diff
@@ -355,6 +382,137 @@ public function save($data) {
...
+                ));
+            }
+
+        }
+
+        if (isset($talk['speakers']) && is_array($talk['speakers'])) {
```
akrabat: This entire section needs thinking about as speaker name isn't unique…

src/models/TalkMapper.php
```diff
@@ -512,11 +670,92 @@ public function thisUserHasAdminOn($talk_id) {
...
+     * @param int $talk_id The identifier of the talk to check
+     *
+     * @return bool true if the user is set as speaker, false otherwise
+     */
+    public function thisUserIsSpeakerOn($talk_id) {
+        if (! isset($this->_request->user_id)) {
+            return false;
+        }
+
+        $speakers = $this->getSpeakers($talk_id);
```
akrabat: Why not do a database query here against speaker_id in talk_speaker?

src/controllers/TalksController.php
heiglandreas: So we'd POST (or PUT) the talk-data and after the talk has successfully been send we'd afterwards POST/PUT/DELETE the speakers of the talk to ```/talks/<talk-id>/speakers/```.

That would move a lot of the logic whom to add/delete/update out of the API.

I quite like the idea! @lornajane, what's your opinion?

src/controllers/TalksController.php
heiglandreas: I'd personally say it's a 401 :wink: But I can change that, no problem.

lornajane:
is this still WIP or is it ready to be reviewed at this point?

heiglandreas:
The code does what it is supposed to. (At least in my VM (-; ) So I'll call it no longer WIP. I will add more frisby tests but not necessarily with this PR

akrabat:
@heiglandreas, 

This is too big for me to handle, so we're going to have to split into separate PRs and merge one by one. That way each PR will be small enough that I can review and merge in a more sensible time frame.

This is how I'd like them split:

1. Refactor of the create talk to TalksController
2. Edit talk end point in Talks Controller
3. Delete talk updates
4. Fetch talk speakers
4. Create talk speaker
5. Edit talk speaker
6. Delete talk speaker

A few general points:

* Please name the controller action methods after what they do, not their HTTP method.
* Don't add new code to one of the "HTTP action" methods in a controller as they already do too much.
* Don't touch handle() methods. They are going to be deleted.
* New code should follow PSR-2 naming conventions - i.e. no underscores in classnames.

Does this make sense?

heiglandreas:
Makes perfect sense! I won't be able to get to it until tomorrow though´:wink:.

Some of the general points are the result of copying the way of coding in the current implementation :wink:. I'll do my best to optimize that!

akrabat:
Thanks. Yes, I thought you'd followed the current style, so highlighted what I'd prefer before you created the new PRs.

lornajane:
Closing pending the promised refactor

PR #189, opened 2015-05-28
Pending events

akrabat:
List pending events using a new filter on the /events collection.
Approve and reject pending events on the /events/{id}/approval end point.

Note that approval sends an email, but this is not tested as the VM currently has a regex issue.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -503,4 +503,37 @@ class EventsController extends ApiController {
             exit;
         }
     }
+
+    /**
+     * Approve a pending event by POSTing to /events/{id}/approval
+     *
+     * The body of this request is completely irrelevant, simply POSTing to this
+     * endpoint is all that's needed to approve an pending event
+     *
+     * @param  Request $request
+     * @param  PDO $db
+     * @return void
+     */
+    public function approveAction($request, $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to create data", 400);
+        }
+        
+        $event_id = $this->getItemId($request);
+        $event_mapper = new EventMapper($db, $request);
+
+        if (!$event_mapper->thisUserCanApproveEvents()) {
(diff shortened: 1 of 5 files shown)


src/controllers/EventsController.php
```diff
@@ -68,9 +68,21 @@ public function getAction($request, $db) {
...
                 // collection type filter
-                $filters = array("hot", "upcoming", "past", "cfp");
+                $filters = array("hot", "upcoming", "past", "cfp", "pending");
                 if(isset($request->parameters['filter']) && in_array($request->parameters['filter'], $filters)) {
                     $params["filter"] = $request->parameters['filter'];
+
+                    // for pending events we need a logged in user with the correct permissions
+                    if ($params["filter"] == 'pending') {
+                        if (!isset($request->user_id)) {
+                            throw new Exception("You must be logged in to view pending events", 400);
```
heiglandreas: Shouldn't this be a 401?

akrabat: No, because with a 401, you have to respond with a WWW-Authenticate header.

src/controllers/EventsController.php
```diff
@@ -68,9 +68,21 @@ public function getAction($request, $db) {
...
+                        }
+                        $user_mapper= new UserMapper($db, $request);
+                        $canApproveEvents = $user_mapper->isSiteAdmin($request->user_id);
+                        if (!$canApproveEvents) {
+                            throw new Exception("You don't have permission to view pending events", 400);
```
heiglandreas: Shouldn't this be a 403?

src/controllers/EventsController.php
heiglandreas: As described in [RFC5849](https://tools.ietf.org/pdf/rfc5849.pdf#page=27) that should work. But it would need some work on the way the exception triggers the response as we'd have to check for the Exception code being 401 or something else. But thats a different issue.

PR #190, opened 2015-05-31
Sign in with twitter

lornajane:
This adds the API components we need for sign in with twitter.  These are endpoints to:
 - get a request token from twitter that the client can use
 - log in a user who comes in from a trusted client with a request token and verification code that twitter successfully exchanges for an access token.

All twitter API access is from the joind.in API, this should be adaptable to work with other clients but has only been tested with web2 (pull request incoming there also).

To get set up: read config.php.dist in the API project, but basically you need to register a twitter application to get a consumer key and secret, then point the callback at /user/twitter-access on your dev platform URL.

To put this live: we need to create new applications on twitter for the test and live APIs to use (we can't use the existing web1 ones as they point to the wrong callback URL)

Warning: contains very funky model layer remodelling.


magicmonkey:
Jenkins, test this please!

magicmonkey:
Jenkins test this please!  Again.

magicmonkey:
Jenkins, test this please!  Again.  Again.

magicmonkey:
^^^ these problems were to do with the version of PHP which phing was using having switched back to the system-installed version (5.4.36-0+deb7u1), whereas we use a custom-installed 5.6.4 on the server.

The other problems were making phing work after I'd switched the server to use the correct PHP binary.

src/controllers/TwitterController.php
```diff
@@ -0,0 +1,112 @@
...
+
+        ]);
+
+        $oauth = new Oauth1([
+            'consumer_key'    => $this->config['twitter']['consumer_key'],
+            'consumer_secret' => $this->config['twitter']['consumer_secret'],
+        ]);
+        $client->getEmitter()->attach($oauth);
+
+        $res = $client->post('oauth/request_token');
```
akrabat: Guzzle throws an exception on error. I'm not sure we should be exposing it directly to our client. Thoughts?

src/controllers/TwitterController.php
```diff
@@ -0,0 +1,112 @@
...
+            $requestTokenMapper = new TwitterRequestTokenMapper($db);
+            // $tokens is instance of TwitterRequestTokenModelCollection
+            $tokens = $requestTokenMapper->create($data['oauth_token'], $data['oauth_token_secret']);
+            $output_list = $tokens->getOutputView($request);
+            header("Location: " . $output_list['twitter_request_tokens'][0]['uri'], NULL, 201);
+            return $output_list;
+        } else {
+
+            error_log("Twitter: no request token (" . $res->getStatusCode() . ": " . $res->getBody() . ")");
+            exit;
```
akrabat: The client gets a 200 with a blank body in this situation. We need to throw an exception with an appropriate status code.

PR #191, opened 2015-06-05
Hotfix/joindin 583

heiglandreas:
This PR adds a check for a lowercase authorization-header in index.php to handle HTTP-Requests that contain a lowercase authorization-header as they are send during the frisby-tests

Fixes #JOINDIN-583


akrabat:
According to RFC 2616, headers are completely case-insensitive, so this should be fixed so that a header of "AuTHorizaTion" would also work.

heiglandreas:
Thanks @akrabat for bringing my attention to ```array_change_key_case```. Didn't know (and didn't need)  that one before :wink:

PR #192, opened 2015-06-08
Implement /facebook/token to log a user in

akrabat:
API half of logging in with Facebook.

We take the verification code from the client, send to Facebook to get an access token.
With the access token, read the user's profile to get their email address. From that,
look up who they are, create them a new access token and return which is converted to
an access token from which we read the user's email address and then attempt to log
them in.

### DIFF
--- a/src/controllers/FacebookController.php
+++ b/src/controllers/FacebookController.php
@@ -16,6 +16,11 @@ class FacebookController extends ApiController
      */
     public function logUserIn($request, $db)
     {
+        if (empty($this->config['facebook']['app_id'])
+            || empty($this->config['facebook']['app_secret'])) {
+            throw new Exception("Cannot login via Facebook", 501);
+        }
+
         $clientId = $request->getParameter('client_id');
         $clientSecret = $request->getParameter('client_secret');
         $this->oauthModel = $request->getOauthModel($db);
--- a/src/controllers/FacebookController.php
+++ b/src/controllers/FacebookController.php
@@ -73,7 +73,8 @@ class FacebookController extends ApiController
             throw new Exception("Could not sign in with Facebook", 403);
         }
 
-        trigger_error('Unexpected Facebook error (' . $res->getStatusCode() . ': ' . $res->getBody() . ')', E_USER_WARNING);
+        trigger_error('Unexpected Facebook error (' . $res->getStatusCode()
+            . ': ' . $res->getBody() . ')', E_USER_WARNING);
         throw new Exception("Unexpected Facebook error", 500);
     }
 }
--- a/src/controllers/FacebookController.php
+++ b/src/controllers/FacebookController.php
@@ -73,6 +73,7 @@ class FacebookController extends ApiController
             throw new Exception("Could not sign in with Facebook", 403);
         }
 
-        throw new Exception("Facebook: error (" . $res->getStatusCode() . ": " . $res->getBody() . ")", 500);
+        trigger_error('Unexpected Facebook error (' . $res->getStatusCode() . ': ' . $res->getBody() . ')', E_USER_WARNING);
+        throw new Exception("Unexpected Facebook error", 500);
     }
 }
--- a/src/controllers/FacebookController.php
+++ b/src/controllers/FacebookController.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * Facebook-specific endpoints live here
+ */
+
+use GuzzleHttp\Client;
+
+class FacebookController extends ApiController
+{
+    public function handle(Request $request, $db)
+    {
+        // really need to not require this to be declared
+    }
+
+    /**
+     * Take the verification code from the client, send to Facebook to get an access token.
+     * With the access token, read the user's profile to get their email address.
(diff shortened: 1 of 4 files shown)


src/controllers/FacebookController.php
```diff
@@ -0,0 +1,73 @@
...
+        if (empty($code = $request->getParameter("code"))) {
+            throw new Exception("The request code must be supplied");
+        }
+
+        // exchange code for access token
+        $client = new Client([
+            'headers' => ['Accept' => 'application/json']
+        ]);
+
+        $res = $client->get('https://graph.facebook.com/v2.3/oauth/access_token', [
```
chrishoult: Is there any particular reason you chose to hit the API directly and not via FB's PHP SDK?

src/controllers/FacebookController.php
```diff
@@ -0,0 +1,73 @@
...
+
+class FacebookController extends ApiController
+{
+    /**
+     * Take the verification code from the client, send to Facebook to get an access token.
+     * With the access token, read the user's profile to get their email address.
+     * From that, look up who they are, create them a new access token and return
+     * the same format as we do when logging in a user
+     */
+    public function logUserIn($request, $db)
```
chrishoult: It might be good to gate this function with a config parameter, to allow its disabling; there are config fields that are required further down that there's no sanity check on, and it will probably be handy to disable this functionality when an installation does not have an associated Facebook app (I'm assuming we're intending Joind.in to be portable)

src/controllers/FacebookController.php
akrabat: Didn't cross my mind to even look for an SDK…

src/controllers/FacebookController.php
```diff
@@ -0,0 +1,73 @@
...
+                $result = $this->oauthModel->createAccessTokenFromTrustedEmail($clientId, $email);
+                if ($result) {
+                    return array('access_token' => $result['access_token'], 'user_uri' => $result['user_uri']);
+                }
+            }
+
+            throw new Exception("Could not sign in with Facebook", 403);
+        }
+
+        throw new Exception("Facebook: error (" . $res->getStatusCode() . ": " . $res->getBody() . ")", 500);
```
chrishoult: This is exposing potentially sensitive information from the GraphAPI to the end user of the API, who is not necessarily a Joind.in client. It would be better to give a generic message, and log the specific error instead.

Perhaps in the future an escalated user permission would be appropriate to include the originating Exception message in the API response.

src/controllers/FacebookController.php
```diff
@@ -0,0 +1,73 @@
...
+            // retrieve email address from Facebook profile
+            $res = $client->get('https://graph.facebook.com/me', [
+                'query' => [
+                    'access_token' => $access_token,
+                ]
+            ]);
+            if ($res->getStatusCode() == 200) {
+                $data = json_decode((string)$res->getBody(), true);
+                if (!array_key_exists('email', $data)) {
+                    throw new Exception("Email address is unavailable", 403);
```
chrishoult: Is this really a 403? It's unlikely to be the caller's fault; I'm not entirely sure what would be appropriate in this instance.

src/models/OAuthModel.php
```diff
@@ -356,4 +356,34 @@ public function createAccessTokenFromTwitterUsername($clientId, $twitterUsername
         return false;
     }
 
+    /**
+     * Create an access token for someone identified by email address via a
+     * third party authentication system such as Facebook
+     *
+     * @param  string $clientId         aka consumer_key (of the joindin client)
+     * @param  string $email            User's email address (that we just got back from authenticating them)
+     * @return string                   access token
```
chrishoult: This method returns either an array or `false`...

src/models/OAuthModel.php
```diff
@@ -356,4 +356,34 @@ public function createAccessTokenFromTwitterUsername($clientId, $twitterUsername
...
+     */
+    public function createAccessTokenFromTrustedEmail($clientId, $email)
```
chrishoult: Does this method need to be public?

src/controllers/FacebookController.php
chrishoult: https://packagist.org/packages/facebook/php-sdk-v4 Here you go, official and maintained and everything :)

src/controllers/FacebookController.php
akrabat: Reading [RFC7231](http://tools.ietf.org/html/rfc7231#section-6.5.3), it felt closest as we're doing an authentication request and refusing to complete the authorisation.

src/controllers/FacebookController.php
akrabat: For just hitting the a couple of endpoints, pulling in the entire Facebook SDK seems a little like overkill to me.

akrabat:
@chrishoult Ready to re-review.

src/controllers/FacebookController.php
chrishoult: A fair enough point - we kind of need a Facebook SDK Lite :)

PR #195, opened 2015-07-11
Psr fixes

jsundquist:
Updated all source files and tests to meet PSR2 standards


lornajane:
Wow, I wasn't expecting this patch!  The API v2 codebase has always been deliberately PSR1 only, we hadn't intended to move it to PSR2.  On one hand, PSR2 would be awesome, on another, this is a huge patch to review (I know that I'm weeks away from being able to get into this) and in the meantime it's a breaking change for all other open and future pull requests.

I think we need to talk about whether we want to try to find a way to accept this, or just admit that we can't handle an unexpected patch this size.  Thoughts, anyone?

akrabat:
I think it makes sense to move to PSR-2. 

However, we need the build.xml updated. We also need to convert to namespaces, remove the files related to current phpcs standard and also fix the rest rest of the errors reported by phpcs.

In terms of timing, I'm happy to update my PRs if this gets in before mine do. On the flip side, if any of the current PRs are merged first, then @jsundquist will need to update this PR.

jsundquist:
@lornajane and @akrabat I would be happy to fix the PR if any other changes make it in before this.  The biggest reason why I did this change is there was a mixture of tabs, spaces, and pure bad formatting through out a number of files.  I ran the project through php_cs_fixer to update all files.  If you would prefer PSR1, it doesn't take much to update the code style to that. 

If your target is PSR1 though, I would highly suggest updating your build.xml to include the php_cs if it does not already to verify that there are no PSR1 errors detected. Looking at your current check styles it only includes PEAR and no reference to PSR1/2.

lornajane:
Closing.  We're unlikely to find time to review and merge such a large unexpected change, and it will just be broken by each additional change to the codebase.  I would be happy to take an update to PSR2 but as mentioned this would need to be complete and include the change to the tasks we use on build to check our standards currently (it's a custom set of sniffs).  Currently, this PR is not in that state.

PR #196, opened 2015-08-09
Move Talk creation into own controller

heiglandreas:
This PR moves creation of a talk into it's own controller.

This PR superseeds PR #186 and is Part 1 of at least 7

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -1,5 +1,11 @@
 [
     {
+        "path": "/events(/[^/]+)/talks/?$",
+        "controller": "TalksController",
+        "action": "createTalkAction",
+        "verbs": ["POST"]
+    },
+    {
         "path": "/events(/[^/]+)*/?$",
         "controller": "EventsController",
         "action": "getAction",
@@ -51,7 +57,7 @@
     {
         "path": "/talks(/[^/]+)*/?$",
         "controller": "TalksController",
-        "action": "postAction",
+        "action": "createTalkAction",
         "verbs": ["POST"]
     },
     {
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -57,7 +57,7 @@
     {
         "path": "/talks(/[^/]+)*/?$",
         "controller": "TalksController",
-        "action": "createTalkAction",
+        "action": "postAction",
         "verbs": ["POST"]
     },
     {
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -150,7 +150,7 @@ class TalksController extends ApiController {
                     throw new Exception("Operation not supported, sorry", 404);
             }
         } else {
-            $this->createTalkAction($request, $db);
+            throw new Exception("method not supported - sorry");
         }
     }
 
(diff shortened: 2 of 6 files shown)


src/controllers/TalksController.php
```diff
@@ -98,21 +103,26 @@ public function postAction($request, $db) {
                                 $this->config['akismet']['apiKey'],
                                 $this->config['akismet']['blog']
                             );
-                            $isValid = $spamCheckService->isCommentAcceptable(
+                            $isValid          = $spamCheckService->isCommentAcceptable(
```
akrabat: whitespace change not needed.

src/config/routes/2.1.json
```diff
@@ -51,7 +57,7 @@
     {
         "path": "/talks(/[^/]+)*/?$",
         "controller": "TalksController",
-        "action": "postAction",
+        "action": "createTalkAction",
```
akrabat: This route needs go to postAction as otherwise, I can't create a comment…

heiglandreas: You're right! Sorry, missed that bit...

src/controllers/TalksController.php
heiglandreas: Shouldn't be there at all... Will be removed!

src/controllers/TalksController.php
```diff
@@ -123,35 +133,179 @@ public function postAction($request, $db) {
...
+     * @return array|bool
+     */
+    public function createTalkAction($request, $db)
+    {
+        // Set the event-ID either via URL or via POST-Data depending on how the
+        // method was called.
+        // When it's called via a POST to /talks/ the event-id is part of the
+        // POST-data, when it's called via POST to /events/<id>/talks/ the
+        // event-id is part of the URL but will be overwritten by POST-data
+        $event_id = null;
```
akrabat: The documentation says:

> To create a talk, you actually need to make a POST request to the *talks* collection of the *event* that the session belongs to

i.e. POSTing to /talks/ doesn't create a new talk. Do we want it to?

src/controllers/TalksController.php
```diff
@@ -150,10 +150,151 @@ public function postAction($request, $db) {
                     throw new Exception("Operation not supported, sorry", 404);
             }
         } else {
-            throw new Exception("method not yet supported - sorry");
+            $this->createTalkAction($request, $db);
```
akrabat: We don't POST to `/talks` to create a new talk. Please put Exception back.

src/inc/Request.php
```diff
@@ -99,6 +99,9 @@ public function __construct($config, array $server, $parseParams = true)
         }
 
         $this->setBase($this->getScheme() . $this->getHost());
+        if (isset($server['HTTP_PORT']) && $server['HTTP_PORT'] && $server['HTTP_PORT'] != 80) {
+            $this->setBase($this->getBase() . ':' . $server['HTTP_PORT']);
+        }
```
akrabat: Why is this part of this PR?

akrabat:
As this PR moves the talk creation code, can you also remove the old code at https://github.com/joindin/joindin-api/blob/master/src/controllers/EventsController.php#L149-L223 please.

Also, the result of PR shouldn't change functionality.

heiglandreas:
I'll split this one up into two:

* Move stuff from EventsController to TalksController (including removal of old code) without changing functionality
* Altering the functionality as it will be needed to comply with editing talks (Might also go into the PR that handles the editing.

As I'd like to include Lornas Removal of the ```handle```-function I think it'S easier to do a new PR based on the current master instead of modifying this one which will become messy.

PR #197, opened 2015-08-16
Feature/edit talksv2 add edit talk endpoint

heiglandreas:
This does NOT implement editing the speakers of a talk as those will be
edited using a dedicated talk_speaker-controller!

This is the second PR replacing #186

This PR is based on PR #196!

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -66,7 +66,12 @@
         "action": "deleteAction",
         "verbs": ["DELETE"]
     },
-
+    {
+        "path": "/talks(/[^/]+)*/?$",
+        "controller": "TalksController",
+        "action": "editTalkAction",
+        "verbs": ["PUT"]
+    },
     {
         "path": "/token(/[^/]+)*/?$",
         "controller": "TokenController",
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -1,5 +1,11 @@
 [
     {
+        "path": "/events(/[^/]+)/talks/?$",
+        "controller": "TalksController",
+        "action": "createTalkAction",
+        "verbs": ["POST"]
+    },
+    {
         "path": "/events(/[^/]+)*/?$",
         "controller": "EventsController",
         "action": "getAction",
@@ -51,7 +57,7 @@
     {
         "path": "/talks(/[^/]+)*/?$",
         "controller": "TalksController",
-        "action": "postAction",
+        "action": "createTalkAction",
         "verbs": ["POST"]
     },
     {
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -1,7 +1,9 @@
 <?php
 
-class TalksController extends ApiController {
-    public function handle(Request $request, $db) {
+class TalksController extends ApiController
+{
+    public function handle(Request $request, $db)
+    {
         if($request->getVerb() == 'GET') {
             return $this->getAction($request, $db);
         } elseif($request->getVerb() == 'POST') {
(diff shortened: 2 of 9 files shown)


src/config/routes/2.1.json
```diff
@@ -1,5 +1,11 @@
 [
     {
+        "path": "/events(/[^/]+)/talks/?$",
```
chrishoult: Should you be catching `\d+` rather than `[^/]+`? Events have numeric IDs; it's probably wise to add a bit of extra sanity

chrishoult: I appreciate that the other routes might not care:)

akrabat:
@chrishoult don't worry about this one or #199 as @heiglandreas has to get the PR related to creating a talk sorted first. See #196.

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+     */
+    public function createTalkAction($request, $db)
+    {
+        // Set the event-ID either via URL or via POST-Data depending on how the
+        // method was called.
+        // When it's called via a POST to /talks/ the event-id is part of the
+        // POST-data, when it's called via POST to /events/<id>/talks/ the
+        // event-id is part of the URL but will be overwritten by POST-data
+        $event_id = null;
+        if ($request->url_elements[2] == 'events' && isset($request->url_elements[3])) {
```
chrishoult: I swear I coded the Routers to pick out named parameters in the URLs; it would be good to drop the `url_elements` part at some point

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+        // incoming data
+        $talk   = array();
+        $errors = array();
+
+        $talk['event_id'] = $event_id;
+        if (empty($talk['event_id'])) {
+            $errors[] = '"event_id" is a required field';
+        }
+
+        $talk['talk_title'] = filter_var(
```
chrishoult: You could probably achieve the below a bit more elegantly with a "schema" of sorts and a loop, eg:

```php
$schema = [
    'event_id' => FILTER_SANITIZE_INT,
    'talk_title' => FILTER_SANITIZE_STRING,
    'talk_description' => FILTER_SANITIZE_STRING,
    'talk_title' => FILTER_SANITIZE_STRING,
    'type' => FILTER_SANITIZE_STRING
];

foreach ($schema as $key => $filter) {
    $talk[$key] = filter_var($request->getParameter($key), $filter);
    if (empty($talk[$key])) {
        $errors[] = "The '{$key}' field is required";
    }
}
```

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+        $event_mapper->cacheTalkCount($talk['event_id']);
+
+        $incoming_speakers_list = (array) $request->getParameter('speakers');
+        $speakers = array_map(function($speaker){
+            $speaker = filter_var($speaker, FILTER_SANITIZE_STRING);
+            $speaker = trim($speaker);
+            return $speaker;
+        }, $incoming_speakers_list);
+
+        foreach($speakers as $speaker) {
```
chrishoult: Is this code out of place? It doesn't seem to have any effect...

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+            'talk_id' => $talk_id,
+            'event_id' => $existing_talk['event_id']
+        );
+
+        $talk['talk_title'] = filter_var(
+            $request->getParameter("talk_title"),
+            FILTER_SANITIZE_STRING
+        );
+        if(empty($talk['talk_title'])) {
+            $errors[] = "'title' is a required field";
```
chrishoult: The error messages here are repeated and are also inconsistent with the create method

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+        $new_talk = $talk_mapper->getTalkById($new_id);
+        return $new_talk;
+    }
+
+    /**
+     * Edit a talk
+     *
+     * This action expects the following parameters:
+     *
+     * - string title
```
chrishoult: These parameters don't match with the code below

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+
+        $isAdmin   = $talk_mapper->thisUserHasAdminOn($talk_id);
+        //$isSpeaker = $talk_mapper->thisUserIsSpeakerOn($talk_id);
+        //if (! $isAdmin  && ! $isSpeaker) {
+        if (! $isAdmin) {
+            throw new Exception('You are not entitled to edit this entry', 403);
+        }
+
+        $eUri = explode('/', $existing_talk['event_uri']);
+        $existing_talk['event_id'] = $eUri[(count($eUri)-1)];
```
chrishoult: Is the talk's `event_id` not coming back with the hit to the mapper?

src/controllers/TalksController.php
```diff
@@ -123,37 +133,307 @@ public function postAction($request, $db) {
...
+            }
         }
+
+        if ($errors) {
+            throw new Exception(implode(". ", $errors), 400);
+        }
+
+        $talk_mapper->edit($talk, $talk_id);
+
+        header("HTTP/1.1 204 No Content", NULL, 204);
```
chrishoult: You can't be guaranteed that HTTP/1.1 was the protocol used to make the request; can you make use of `http_response_code` instead?

akrabat: Traditionally, we do:

    header("Content-Length: 0", NULL, 204)

c.f: https://github.com/joindin/joindin-api/blob/master/src/controllers/UsersController.php#L71

I'd argue we should do the same here for this PR and if it's wrong, then change it to what's correct in a separate PR.

src/models/TalkMapper.php
```diff
@@ -313,48 +313,177 @@ public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose =
...
+     * - duration
+     * - speakers (an array of names)
+     * - type (a value from the column categories:title)
+     *
+     * @param array $talk    talk data to insert into the database.
+     * @param int   $talk_id The ID of the talk to be edited
+     *
+     * @return integer|false
+     */
+    public function edit($talk, $talk_id)
```
chrishoult: You probably want to type-hint `$talk` as an `array`

chrishoult:
LOL got a little trigger happy on the PR :)

@heiglandreas A general note: I'd be wary of reformatting existing code just for the sake of it; this PR introduced line-breaks in code that you weren't otherwise altering.

src/controllers/TalksController.php
chrishoult: Ah yeah of course - what @akrabat said :)

heiglandreas:
Thanks for all the comments! I close this PR as it will not merge with the new Series of PRs that will hopefully be easier to digest ;)

PR #198, opened 2015-08-22
Remove handle() method from controllers

lornajane:
The previous routing strategy relied on verb routing inside the controllers; it is no longer needed but the parent controller class implemented the handle() method as an abstract so it always had to be declared.  The requirement to have it required is gone and the existing handle methods are also removed


heiglandreas:
:+1:

PR #199, opened 2015-08-23
Edit talks v2: fetch speakers

heiglandreas:
This PR allows to fetch already available speakers for a talk. It introduces 2 API-Endpoints:

* /v2.1/talks/[talk-id]/speakers will give a list of speakers for the talk [talk-id] with references to the user-url if the talk has been claimed
* /v2.1/speakers/[speaker-id] will give a list with one entry for the speaker [speaker-id].

Both results look like the following JSON where the list for the **/v2.1/talks/[talk-id]/speakers** can contain multiple entries and the list for the **/v2.1/speakers/[speaker-id]** will only contain 1 entry (the entry for the requested speaker-entry).

Depending on whether the user has claimed the talk the ```full_name``` and ```twitter_username``` entries will contain a value or not and if the talk is claimed the entry ```user_uri``` is available or not.


src/models/TalkSpeakerMapper.php
```diff
@@ -0,0 +1,162 @@
...
+    public function getDefaultFields() {
+        $fields = array(
+            "username" => "speaker_name",
+            "full_name" => "full_name",
+            "twitter_username" => "twitter_username"
+        );
+        return $fields;
+    }
+
+    public function getVerboseFields() {
```
chrishoult: Would it not be more future-proof to return `$this->getDefaultFields();`?

src/models/TalkSpeakerMapper.php
```diff
@@ -0,0 +1,162 @@
...
+
+        // group by because of adding the user_attend join
+        $sql .= ' group by speaker.ID ';
+
+        // order by
+        if ($order) {
+            $sql .= ' order by ' . $order;
+        }
+
+        error_log($sql);
```
chrishoult: Is this left in intentionally? If so, you're missing more SQL below :)

src/models/TalkSpeakerMapper.php
```diff
@@ -0,0 +1,162 @@
...
+
+        $countStmt = $this->_db->prepare($sql);
```
chrishoult: This isn't used...

akrabat:
I'm closing this as the talk claims issue [JOINDIN-637](https://joindin.jira.com/browse/JOINDIN-637) overlaps.

PR #200, opened 2015-08-25
Move talk-creation to TalkController

heiglandreas:
This PR moves talk creation from the Events-Controller to the Talks
controller and adapts the routing appropriately.

Additionally tests for minimum-talk creation have been added to make
sure talk-creation works as expected.

This is the first of a series of PRs that will follow to enable talk-editing via API and addresses issue JOINDIN-525

PR #201, opened 2015-08-28
alter talk creation to include all fields

heiglandreas:
This PR is based on PR #200 and should first be revised after that one is merged as the differences are then easier to see.

This PR adds two more fields (duration and talk-type) to newly created talks. It also changes handling of some of the old fields and tries to optimize error-handling by showing the user all errors at once instead of only showing the first error.

It also introduces a new ```isLanguageValid()```-method to the LanguageMapper.


akrabat:
Looks like we're waiting for this to be rebased?

heiglandreas:
Rebased

akrabat:
As far as I can tell, this PR does:

1. Adds ability to retreive list of talk categories (talk types) from the database
2. Adds ability to add a new talk catetory (talk type) to the database
3. Add "duration" parameter
4. Add slides_link" parameter
5. Renames "talk_type" parameter to "type"
6. Changes "type" to be a required field
7. Validate "type" parameter against list of categories in `talk_cat` table
8. Validate "language" parameter against list of languages in `lang` table
9. Rewords error message when talk title field is missing
10. Rewords error message when talk description field is missing
11. Rewords error message when start date field is missing
12. Adds ability to set the talk type when creating a talk as previously it was hardcoded to 1 ("Talk"). This is implemented as a separate TalkMapper method which also ensures that the talk can only belong in one category.
13. Moves adding of speakers to a talk from the TalkMapper's createTalk() to the TalkController's createTalkAction().
14. Updates frisky write talks tests for reworded error messages
15. Adds new frisky write talks test for newly required talk type field
16. Breaks the Frisby write test testCreateApprovedEvent()

Notes:

* 5 & 6 are BC breaks with the published documentation (http://joindin.github.io/joindin-api/talks.html#creating-new-talks)
* 9, 10 & 11 are technically BC breaks


Don't make any changes to this PR please as I need to discuss with @lornajane.

heiglandreas:
Undoing 5, 6, 9, 10 and 11 is no problem at all.

PR #203, opened 2015-09-06
JOINDIN-579 - API for retrieving list of talk types (categories)

dstockto:
GET .../v2.1/categories
GET .../v2.1/categories/<ID>

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -180,6 +180,20 @@
     },
 
     {
+        "path": "/categories/?$",
+        "controller": "CategoriesController",
+        "action": "getAllCategories",
+        "verbs": ["GET"]
+    },
+
+    {
+        "path": "/categories(/[0-9]+)?/?$",
+        "controller": "CategoriesController",
+        "action": "getCategory",
+        "verbs": ["GET"]
+    },
+
+    {
         "path": "/?$",
         "controller": "DefaultController",
         "action": "handle"
--- a/src/controllers/CategoriesController.php
+++ b/src/controllers/CategoriesController.php
@@ -1,0 +1,32 @@
+<?php
+class CategoriesController extends ApiController
+{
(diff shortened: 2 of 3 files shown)


src/config/routes/2.1.json
```diff
@@ -174,6 +174,20 @@
     },
 
     {
+        "path": "/categories$",
```
chrishoult: Most of the other routes have `/?` at the end to cope with a trailing slash or not; please add this in.

dstockto: Will do.

dstockto:
@chrishoult Updated to allow for optional trailing slash

src/controllers/DefaultController.php
```diff
@@ -8,6 +8,7 @@ public function handle(Request $request, $db)
 
         // just add the available methods, with links
         $retval['events']          = $request->base . '/' . $request->version . '/events';
+        $retval['categories']      = $request->base . '/' . $request->version . '/categories';
```
lornajane: I'm not sure how useful it is to offer this link at the entry level of the API, I mostly think of this as a list of places a user wants to go

dstockto: I'm happy to remove the entry and resubmit the PR if you'd like. On the other hand, do we have a place for a developer to go who wants to integrate w/ joind.in and needs to know what APIs are available?

lornajane:
We do have docs, which should include this, but that's probably the only place it would be

dstockto:
I'll remove the entry and resubmit then. Thank you

akrabat:
The `/languages' endpoint has `name` and this one has `title` for the same piece of data. Should we be consistent and make `/categories` also use `name`?

lornajane:
Merged but if you wanted to write a test to go with it, that would also be fine ;)

PR #211, opened 2015-09-06
Codesniff - Update TalksController to PSR-2

dstockto:
No functionality, just style.

One missing update is that the switch statement is supposed to have a comment if there's a fall-through. I don't know if that fall-through is intentional or a bug. As such, I've not fixed that codesniff error.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -1,39 +1,41 @@
 <?php
 
-class TalksController extends ApiController {
-	public function getAction($request, $db) {
+class TalksController extends ApiController
+{
+    public function getAction($request, $db)
+    {
         $talk_id = $this->getItemId($request);
 
         // verbosity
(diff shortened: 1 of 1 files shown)


akrabat:
The CS error is because of the else on line [126](https://github.com/joindin/joindin-api/blob/master/src/controllers/TalksController.php#L126-L128).

The else isn't needed, but I'm not sure if phpcs knows that throwing an exception before the next `case` means that there is no fall-through.

dstockto:
In that case, what would you prefer? I can add a break which would make it shut up but would be unreachable and not needed technically. Or we can ignore it.

lornajane:
I think the unnecessary break statement gets my vote, it's expected by humans as well as static analysis tools so it adds either nothing or readability, probably just depending on the reader.

dstockto:
I'll update shortly.

PR #239, opened 2015-09-07
JOINDIN-590 - Increase range of duplicate inflected titles from 5 to 9

dstockto:
When the API is used to retrieve a talk that doesn't have any value in
the url_friendly_talk_title, it attempts to create a value for the
field. If there are multiple talks with the same title for the same
event, it will try over and over to append a new number at the end. It
will stop after attempting suffixes 2 through 9 if none of those work.
It should be Good Enough™ for now.

### DIFF
--- a/src/models/TalkMapper.php
+++ b/src/models/TalkMapper.php
@@ -443,22 +443,21 @@ class TalkMapper extends ApiMapper {
      */
     protected function generateInflectedTitle($title, $talk_id)
     {
+        // Try to store without a suffix
         $inflected_title = $this->inflect($title);
         $result = $this->storeInflectedTitle($inflected_title, $talk_id);
         if ($result) {
             return $inflected_title;
         }
-        
-        // Add a number to the
-        $inflected_title .= '-1';
-        for ($i=2; $i <=5; $i++) {
-            $inflected_title = preg_replace('/-(\d)$/', "-$i", $inflected_title);
-            $result = $this->storeInflectedTitle($inflected_title, $talk_id);
-            if ($result) {
-                return $inflected_title;
-            }
+
+        // If that doesn't work, try to store with the talk ID tacked on
+        $inflected_title .= '-' . $talk_id;
+        $result = $this->storeInflectedTitle($inflected_title, $talk_id);
+        if ($result) {
+            return $inflected_title;
         }
 
+        // If neither of those worked, fail. We shouldn't get to here.
         return false;
     }
 


akrabat:
Out of interest, did you actually come across a situation where 4 attempts wasn't enough?

dstockto:
@akrabat In the links Larry gave in the ticket in JIRA there were a few where there are 5 or 6 of the same names during a conference. The names are numbered in the order they are retrieved. The ones that are beyond the -5 suffix are the ones that return false for that field.

heiglandreas:
Just out of interest: Wouldn't appending a random number out of a much wider range be a faster way? Or for the sake of it a random letter? Or is there a need for the sequence?

dstockto:
@heiglandreas In theory, yes. Appending a random letter or larger number would make the chances of collisions less, but it would also appear to be more "random" looking. Right now, the first talk that is fetched receives a non augmented name. The subsequent ones will get -2 through -9 in the order they are retrieved. But even with this, there's still the chance of collision.

This change was made to be the least impactful, right now, to what was happening and to hopefully cover the few cases we do have where 5 was not enough to cover the talks with the same name during the same event. I'd look at it as a stop-gap measure with a better solution coming sometime in the future.

chrishoult:
Might be a little ignorant, but would appending the Talk's ID not be a much more reliable solution?

akrabat:
It shouldn't be a real problem for real data as the talk only needs to be unique per event.

dstockto:
@chrishoult Yes, I think that would be perfect. Talk IDs are unique in the system, so it would presumably never need to be retried. There are theoretical ways we could make this fail but it's not likely to ever happen in real world data.

@akrabat The catalyst for this ticket was real data. Larry was finding that there were several talks which don't have the URL friendly talk title field.

Are there any objections to removing the retry functionality and just doing @chrishoult's suggestion of appending the talk ID? It's still doing updates to data on a GET (ugh) but at least it's not trying (up to) 9 times to do it.

/cc @lornajane

lornajane:
Yeah sorry about the GET, it's for data entered via web1 and it should only happen once... refactoring welcome :)  Agree that this is a real problem and that the talk ID is a perfectly excellent way of solving it

dstockto:
On it.

dstockto:
For talks that are not duplicated (or the first one hit) should we store the inflected title without any numerical decorations? Or should we always tack on the id for any that are being updated?

If it fails to write for some reason (not sure what that would be, but let's assume there's already a talk in the event that has the same url_friendly_talk_title, should we then go to tacking on numbers, or consider it failed at that point?

/cc @lornajane @chrishoult @akrabat

dstockto:
Updated to try the undecorated version of the inflected title first. If that doesn't work it will try to put in the talk ID. If that still doesn't work, it will fail.

akrabat:
Have to admit that I preferred the incremented number as it looks nicer in a URL :)

However, this does work and I spot-checked a few of the endpoints listed in the issue and they now have valid `url_friendly_talk_title`s

PR #257, opened 2015-09-11
Update PHPCS rules to a variant of PSR2

akrabat:
Use PSR2 without the namespace rule as we don't have namespaces in -api yet.


akrabat:
Ironically, our codebase is closer to PSR2 than it is to the original JoindIn codesniffer rules :)

PR #259, opened 2015-09-16
Update frisby write tests for talks

akrabat:
**Note** This PR is required to make the `API_v2_run_destructive_tests` task on Jenkins work again.

The frisby write tests need to use an access token for a user that we've created during the run. To avoid creating another user, we use the access token from the event tests and call runTalkTests() after we've retrieved a newly created event.

This has the benefit that we can refactor the talk tests to use the event's talks_uri property and so the talks tests do not need to manually construct the URI to post to.

Finally, as the talks tests are now called within api_write_events.js, we don't need to explicitly call them api_write_spec and can remove the set up code from api_write_talks.js.

### DIFF
--- a/tests/frisby/api_write_events.js
+++ b/tests/frisby/api_write_events.js
@@ -2,6 +2,7 @@
 
 var frisby   = require('frisby');
 var datatest = require('./data');
+var talkstest  = require('./api_write_talks');
 var util     = require('util');
 
 var baseURL = '';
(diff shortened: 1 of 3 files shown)


heiglandreas:
That looks much more sensible!

heiglandreas:
From my side, that's OK. Thanks @akrabat for fixing and making the test much more sensible!

PR #264, opened 2015-10-02
JOINDIN-600: Comment poster should not get comment notification

bcremer:


phoenixrises:
Is the additional function really required?

bcremer:
@phoenixrises it's not required but it's private and in my opinion it makes the intent of the condition clearer.

phoenixrises:
But you could say that for just about any condition ever couldn't you? Personally a comment would have done imo, but will leave it up to the reviewers, either way, at least it's done :)

akrabat:
ok to test

akrabat:
This requires more work as there's no guarantee that `$person` will have a `user_id` column (unclaimed speakers).

I'm closing for now. Feel free to reopen when addressed.

PR #265, opened 2015-10-02
Added talk_link_types table

toby-griffiths:
As requested by Rob

### DIFF
--- a/db/patch60.sql
+++ b/db/patch60.sql
@@ -1,0 +1,12 @@
+-- Add link types table
+
+CREATE TABLE `talk_link_types` (
+  `ID`   INT(11)      NOT NULL AUTO_INCREMENT,
+  `type` VARCHAR(255) NOT NULL DEFAULT '',
+  PRIMARY KEY (`ID`)
+)
+  ENGINE = InnoDB
+  DEFAULT CHARSET = utf8;
+
+INSERT INTO patch_history
+SET patch_number = 60;


lornajane:
add to whitelist

akrabat:
bah! Can't type a number. Sorry.

lucianoqueiroz:
Do we need some seed for this table? What types of link will be available?

heiglandreas:
I'd at least seed it with 'slides' and perhaps 'video' would be another option. Further suggestions?

lucianoqueiroz:
We already have the "slides_link" column on talks table. The description on Jira says: _Often talks have items in addition to slides - maybe a zip file of code examples or a podcast or a video or something - the ability to add additional links to the talk with this information would be great.
_

lucianoqueiroz:
I think we can have: Video, Code Examples, Podcast, Text ?

heiglandreas:
Will we put the slides in there as well? Or shall they remain in the talk-table? I'd move them over as well. So there would then also be 'Slides'

akrabat:
> Will we put the slides in there as well? Or shall they remain in the talk-table? I'd move them over as well. So there would then also be 'Slides'

Legacy uses the slides column in that table. I seem to remember discussion that we'd put slides in the new table, and automatically sync to the column in the talks table.

akrabat:
Found it!

See https://joindin.jira.com/browse/JOINDIN-650 & https://joindin.jira.com/browse/JOINDIN-651

lucianoqueiroz:
> The old website expects a slides_link field in the talks table. We now store this information in the new talk_links table, but we need to manage the migration here.
> When a talk_link is created, add the URL to the talks table if there is not already a URL there.
> When a talk_link is edited and the URL changes, check if the same URL was stored in the talk.slides_link column and if so, update that also.
> When a talk_link is deleted, if its URL is used in the talks table, empty that field. If there are other valid talk_links for this talk, add one to the talks table.

As we will have multiple links per talk, what if the user first adds other link rather than "slides"? The label on legacy site expects a slide url...

akrabat:
This conversation should be held on the relevant JIRA ticket as it will be lost here.

PR #266, opened 2015-10-02
JOINDIN-496  - Added extra clause to SQL query to support searching s…

iaini:
…tub shortcode from search box
Does not completely solve issue JOINDIN-496, but might help


lornajane:
ok to test

akrabat:
I'm really sorry, but I'm not going to take this. Having thought some more about it, I'd rather wait for the elastic search integration.

PR #268, opened 2015-10-07
Do not HTML encode quotes

walterebert:
Added flag FILTER_FLAG_NO_ENCODE_QUOTES to FILTER_SANITIZE_STRING to prevent encoding of quotes, because Twig will double escape them.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -89,12 +89,12 @@ class EventsController extends ApiController
                 }
 
                 if (isset($request->parameters['title'])) {
-                    $title           = filter_var($request->parameters['title'], FILTER_SANITIZE_STRING);
+                    $title           = filter_var($request->parameters['title'], FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES);
                     $params["title"] = $title;
                 }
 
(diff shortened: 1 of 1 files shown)


dstockto:
ok to test

akrabat:
I think that this is the right approach for new data arriving via the API as HTML output isn't necessarily the only output, so it makes sense to store without quoting.

I've tested the change with web2 and it does what I would expect and stores non encoded quotes to the database.

@richsage: Are there any implications for the iPhone or Android apps?
@heiglandreas, @AurelioDeRosa: Are there any implications for CallingAllPapers.com?

richsage:
@akrabat I'll check here and update accordingly

heiglandreas:
I'll check as well

heiglandreas:
@akrabat I can't see any problems with CallingAllPapers.com, so OK from my side.

@walterebert Any chance for tests?

AurelioDeRosa:
I don't think it'll impact in any way CallingAllPapers.com, so that should be good to merge.

walterebert:
@heiglandreas I haven't used the JS tests yet, so I'll have to look into that first

heiglandreas:
@walterebert Ping me if you have questions :wink:

PR #271, opened 2015-10-31
Location is a required field when submitting events.

pavlakis:


akrabat:
add to whitelist

lornajane:
ok to test

akrabat:
No tests on the docs site :) Clearly, we both have the same reflex when we see Jenkins' message!

PR #276, opened 2015-11-16
Rename categories to talk_types

akrabat:
Although the table in the database is called categories, we refer to
them as talk types. Ensure that throughout the code and the endpoints we
use talk type everywhere.


dstockto:
test this please

PR #278, opened 2015-11-19
Build should fail if codesniff errors or warnings are detected.

dstockto:
Test PR should fail if there are any errors or warnings from codesniffer.


dstockto:
test this please

PR #279, opened 2015-11-19
Enable PDO exceptions

akrabat:
Ideally, we'd log these too, but at a minimum, we shouldn't silently fail when we have a database error.


dstockto:
test this please

PR #286, opened 2015-12-06
Refactor TalkMapper to introduce TalkModelCollection and TalkModel

akrabat:
Implement TalkModelCollection and TalkModel to separate the output representation from the mapper.

This is a fairly large change-set, so I've tried to explain my thinking in each commit message…

Key things:

1. I've introduced `AbstractModel` and `AbstractModelCollection` to hold methods common to the respective `TwitterRequest` and `Talk` classes.
2. The `AbstractModel` has the concept of sub-resources in additional to default and verbose fields so that I could handle speakers and tracks.
3. The `TalkMapper` still needs to pre-process the results from a database read before passing to the Model. This was done in `transformResults()` which is inherited from `ApiMapper`, so I've created a new method called `processResults()` to do this work.
4. The TalkMapper's `getTalkById()` returns a `TalkModel` rather than a collection containing one item. This makes more more sense as it's used by the `TalkCommentEmailService`, which now uses the 'internal representation' of the talk rather than the output view one. This change also meant that the `TalkModelCollection`'s constructor can accept data that's already hydrated.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -17,7 +17,8 @@ class EventsController extends ApiController
             switch ($request->url_elements[4]) {
                 case 'talks':
                     $talk_mapper = new TalkMapper($db, $request);
-                    $list        = $talk_mapper->getTalksByEventId($event_id, $resultsperpage, $start, $verbose);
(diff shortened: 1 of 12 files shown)


akrabat:
test this please

akrabat:
test this please

PR #289, opened 2015-12-13
Add ability to upload an event icon

akrabat:
Implement a new endpoint of /events/{id}/icon which accepts a PUT request to set the icon for that event. The icon must be square and if it one side is larger than 1440 pixels, then it is resized to 1440px.

Also add the event's `icon_uri` to its verbose output.

This is the API part of [JOINDIN-530](https://joindin.jira.com/browse/JOINDIN-530).

### DIFF
--- a/src/config.php.dist
+++ b/src/config.php.dist
@@ -6,6 +6,9 @@ $config =  array(
     // the URL of web2 website is used in links within emails
     'website_url' => 'https://m.joind.in',
 
+    // The directory where event icons are stored
+    'event_icon_path' => __DIR__ . '/../../joind.in/src/inc/img/event_icons/',
+
     'oauth' => array(
         'expirable_client_ids' => array(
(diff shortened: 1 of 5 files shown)


lornajane:
The config has been set on the test and live platforms for this PR

lornajane:
Given that many of our users are at events and might be on a poor connection or using mobile data, it would be better to store just the size of image that we need (I think 140px?  So we can do the 70px images that web2 uses nicely for hi-res things).  Alternatively, we could store a larger image and also create our usual size one so that is stored and can be served as well as the bigger one.

akrabat:
I've updated the branch to create two images: a large one and a small one. I've also refactored the image creation code into a Model class as the controller action was becoming unwieldy.

lornajane:
Closing in favour of https://github.com/joindin/joindin-api/pull/315

PR #291, opened 2015-12-15
Ensure that website links are correct

akrabat:
All links to the website in resource should point to the correct instance. 

This is more invasive than I expected as the mappers do the transformation and don't have access to the config.


akrabat:
Replaced by #294.

PR #292, opened 2015-12-15
Add icon_uri to the event resource

akrabat:
This provides `icon_ui` which is the fully qualified URL to the event's icon. It assumes that the image is stored in such a way that it can be accessed via `{website uri}/inc/img/event_icons/{icon filename}`.
 
Note that this PR builds on #291 as it needs the same config variable in the mapper.


akrabat:
Closing and re-implementing as the `EventMapper` now has a `website_url` property.

PR #295, opened 2015-12-30
"all" filter for events collection

akrabat:
The "all" filter provides all active events in ascending date order. If the "start" parameter is not provided, then the collection will provide the first page where there is an event in the future.

This allows a client to go "backwards" from this point to retrieve past events seamlessly.

### DIFF
--- a/src/controllers/ApiController.php
+++ b/src/controllers/ApiController.php
@@ -39,7 +39,7 @@ abstract class ApiController
 
     public function getStart($request)
     {
-        return (int) $request->paginationParameters['start'];
+        return $request->paginationParameters['start'];
 
     }
 
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -72,7 +72,7 @@ class EventsController extends ApiController
                 $params = array();
 
                 // collection type filter
-                $filters = array("hot", "upcoming", "past", "cfp", "pending");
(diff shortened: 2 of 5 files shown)


lornajane:
I *love* this feature but I'm not sure about the interface.  Actually now I've fiddled with it a bit more the only thing that still seems weird is:
 - the event you're looking for (i.e. the soonest-future event) should be the first event in the default collection page returned.  So we should set start to the right number rather than adjusting for page size so that if you want to retrieve the next events and go backwards/forwards, you don't have to get one dataset and then filter out to try to work out which are in the future and which are in the past.  Strangely-sized pages at the extremes is less of a problem IMO.

akrabat:
I started with what you suggested, but I can't work out how to make it work within web2 when the url is of the format `https://joind.in/event?page=2` with the next page of `https://joind.in/event?page=3`.

The only way I could think to do it was to change the URLs on web2 to expose the way the API works and have a URL such as: `https://joind.in/event?start=32`. That seems like a weird URL to me though as the next page would then be `https://joind.in/event?start=42`. 

May be we don't care about that though?

akrabat:
Updated to leave `start` in the right place.

lornajane:
Can we store where the "middle" is internally and expose page=1, page=2 and then page=-1 when going backwards in time?

PR #299, opened 2016-01-02
Remove 'speaker' column from generated talk SQL.

richsage:
This is handled by the talk_speaker table now.


lornajane:
Oops, we were too slow and got this PR twice.  I merged https://github.com/joindin/joindin-api/pull/303

PR #300, opened 2016-01-04
Exclude private events from pending events count

lornajane:
The email notifications show an inaccurate count of events pending
because they weren't excluding the private events, which we won't
approve anyway

### DIFF
--- a/src/models/EventMapper.php
+++ b/src/models/EventMapper.php
@@ -972,7 +972,7 @@ class EventMapper extends ApiMapper
     {
         $sql = 'select count(*) as count '
                . 'from events '
-               . 'where pending = 1';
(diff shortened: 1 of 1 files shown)


lornajane:
Ah this didn't work quite as I expected, I'll revisit

akrabat:
Testing locally, `private` is a `varchar` and reading the web1 source code, it's supposed to be either 'Y' or 'N'. However, api sets it to '0' on submit. It's also a nullable field, hence a private event is "Y" and a public one is either "N", "0" or NULL.

This explains why elsewhere in api, we use `and (events.private <> "y" OR events.private IS NULL)`. (Except for one case where we don't check for NULL…)

PR #301, opened 2016-01-06
Adds a way to retrieve speakers for a talk

heiglandreas:
This PR adds a way to get the speaker(s) for a talk via the API. That way a list of speakers can be retrieved for a single talk without having to fetch the complete talk informations.

It is also a prerequisite to add, edit or delete speakers for a talk which in turn is a prerequisite to edit talks.

This commit introduces two new API-Endpoints:

* /talks(/[^/]+)*/speakers/?$ to get a list of all the speakers for a single talk and
* /speakers(/[^/]+)*/?$ to get information about a single speaker. This endpoint currently returns an array always containing a single entry. That's the same behaviour as the /talks(/[^/]+)*$-endpoint shows.

A speaker can be either a name if the speaker didn't claim the talk or a user when the speaker did claim the talk. So there might be a user-URI on a speaker-entry but that depends on whether the user claimed the talk or not. And a speaker is unique to the talk. So one user doing two talks can be found with two different speaker-entries.

There currently aren't frisby-tests. I'll add them in a follow-up PR later.

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -60,6 +60,12 @@
         "verbs": ["DELETE"]
     },
     {
+        "path": "/talks(/[^/]+)*/speakers/?$",
+        "controller": "Talk_speakersController",
+        "action": "listSpeakersAction",
+        "verbs": ["GET"]
+    },
+    {
         "path": "/talks(/[^/]+)*/?$",
         "controller": "TalksController",
         "action": "getAction",
@@ -77,7 +83,12 @@
         "action": "deleteAction",
         "verbs": ["DELETE"]
     },
-
+    {
+        "path": "/speakers(/[^/]+)*/?$",
+        "controller": "Talk_speakersController",
+        "action": "getSpeakerAction",
+        "verbs" : ["GET"]
+    },
     {
         "path": "/token(/[^/]+)*/?$",
         "controller": "TokenController",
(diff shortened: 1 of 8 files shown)


lornajane:
What's the use case for this additional endpoint?  I'm struggling to see a circumstance where I'd want the speakers but NOT the talk they gave.

heiglandreas:
It's a sideeffect for editing talks. On editing talks it becomes rather tricky to check whether the speakers have changed or a speaker has been added or removed or edited. So I thought it would be an idea to edit the talk (but not the speakers) and edit speakers separately. To be able to do that I thought starting with retrieving speakers for an event would be a good idea to introduce the Controller and Mapper as it doesn't break anything, just adds something. 

I wanted to introduce manipulating speakers for the talks *before* actually editing the talks so when editing the talks would be possible we could edit *all* aspects of the talk including the speakers.

Does that make sense? Or is it complete nonsens? If so, feel free to close the PR :wink:

lornajane:
That makes more sense but I haven't thought about how to handle editing nested records.  Do we have this in place anywhere else?  Like events and tracks/tags/something?

lornajane:
Can we start by just editing talks and making speakers an array of strings containing the speakers' names - like what happens on web1 before a speaker has claimed their talk?  I'd really really like to have add and edit talks, and the talk claims adds a lot of complexity that I don't think we need immediately.  To that end, I'll close this PR but is there an issue that I should be attaching these ideas to?  Can you either update it or point me at it please?

PR #305, opened 2016-01-11
Removes CfPs that have not yet started

heiglandreas:
Currently CfPs that have not yet started will be shown in the CfP-list. This PR fixes that by removing CfPs with start-date greater than the current date from the list. That's the same behaviour as removing CfPs that are already closed from the CfP-List

### DIFF
--- a/src/models/EventMapper.php
+++ b/src/models/EventMapper.php
@@ -166,8 +166,9 @@ class EventMapper extends ApiMapper
                     $where .= sprintf(
                         ' AND events.event_cfp_url IS NOT NULL' .
                         ' AND events.event_cfp_end >= %1$d' .
-                        ' AND events.event_cfp_start <= %1$s',
-                        mktime(0, 0, 0)
+                        ' AND events.event_cfp_start <= %2$s',
+                        mktime(0, 0, 0),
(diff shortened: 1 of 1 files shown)


dstockto:
It seems it might be valuable to know what CfPs are about to start.

heiglandreas:
It might happen (and already happened) that an event was added to joind.in that had a CfP start in 2 months time. So that CfP will be shown for 2 months without it being open. And as the Web-Interface didn't show the beginning or the end of a CfP in the CfP sidebar or on the CfP-Page I would assume as a user that those CfPs are already open.

That was the reason why I removed the CfPs that had not yet started from that list.

If we decide otherwise I'm fine (but will have to adapt callingallpapers accordingly :wink: )

akrabat:
Maybe show those that will start within the next 7 days?

heiglandreas:
No problem. I'll fix that ASAP

heiglandreas:
I've just seen that in the API the key to the CfPs is labeled "open-cfps". (http://api.joind.in) - So should we reconsider? I've just implemented the "7 days in advance".

lornajane:
Looks good to me, thanks Andreas :)

PR #307, opened 2016-01-24
Add icon_uri to the event resource

akrabat:
This provides `icon_ui` which is the fully qualified URL to the event's icon. The URL to the icon image is a config variable `event_icon_url` so that we can support changing it at a later date.

(This PR will probably conflict in config.php.dist after #289 is merged…)


lornajane:
Covered by #315

PR #314, opened 2016-02-28
[Docs] Remove pygments CSS file

akrabat:
GitHub pages has moved to Jekyll 3.0 with the Rouge highlighter. The pygments_trac.css file makes the code unreadable, so while we now do not have any syntax highlighting, at least you can read it!

See http://joindin.github.io/joindin-api/events.html#editing-events for a good example.


lornajane:
Thanks for sorting this, I saw it but had no idea how to fix as my local copy didn't do it.  Local Jekyll all updated and both versions looking better now!

PR #318, opened 2016-03-07
Convert move_uploaded_file warnings to exceptions

akrabat:
This ensures that we send back an error to the client.


lornajane:
What causes the warnings?  Is there a particular use case that I can look at to see this behaviour and therefore test the patch, please?

akrabat:
This happens when the first upload of a logo is stored with permissions such that the second upload doesn't allow it to be overwritten.

akrabat:
This is the -vm PR that solves the fundamental problem: https://github.com/joindin/joindin-vm/pull/85

lornajane:
If we've solved the actual, dev-platform-only problem, do we need a code solution as well?  I'm not sure what this adds that we'd find useful other than on an old dev platform the error would be visible maybe?  Which is to say, I'm not in favour on this one ...

akrabat:
I'm generally against silent failures, but am not precious about this, so feel free to close.

PR #321, opened 2016-03-25
Edit and delete track

akrabat:
PUT to the track's URI to edit the track.
DELETE to the track's URI to delete the track.

PR #323, opened 2016-03-26
Add talk to track

akrabat:
POST to the track URI with a `talk_uri` in the body will add this talk to the track. Checks are in place to ensure that the talk and track both belong to the same event.
 
Note that this PR builds on #321 (Edit track) as it uses the `getEventByTrackId()` method from that PR.


akrabat:
Rebased against updated edit-track branch

akrabat:
This is the wrong way to do it. Need to POST to a `tracks` collection on the talk.

PR #326, opened 2016-03-26
Edit talk

akrabat:
PUT to the talk's `uri`.

Note that this PR builds on [#320 (Add speakers when creating a talk )](https://github.com/joindin/joindin-api/pull/320) as it refactors `addSpeakersToTalk()` into `updateSpeakersOnTalk()`.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -228,8 +228,8 @@ class TalksController extends ApiController
             throw new Exception("You must be logged in to create data", 400);
         }
 
-        $talk['event_id'] = $this->getItemId($request);
-        if (empty($talk['event_id'])) {
+        $event_id = $this->getItemId($request);
+        if (empty($event_id)) {
             throw new Exception(
                 "POST expects a talk representation sent to a specific event URL",
                 400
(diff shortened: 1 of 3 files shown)


src/controllers/TalksController.php
```diff
@@ -238,20 +238,97 @@ public function createTalkAction(Request $request, PDO $db)
...
+    /**
+     * Edit a talk
+     *
+     * Edit talk after being called via the URL "/talks/[talkId]"
+     *
+     * @param Request $request
+     * @param PDO     $db
+     *
+     * @throws Exception
+     * @return array|bool
```
heiglandreas: This method doesn't return anything.

src/controllers/TalksController.php
```diff
@@ -238,20 +238,97 @@ public function createTalkAction(Request $request, PDO $db)
...
+
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
+        if (!($is_admin || $is_speaker)) {
+            throw new Exception("You do not have permission to update this talk", 400);
```
heiglandreas: This should be a 403

src/controllers/TalksController.php
```diff
@@ -238,20 +238,97 @@ public function createTalkAction(Request $request, PDO $db)
...
+        }
+
+        // retrieve the talk data from the request
+        $data = $this->getTalkDataFromRequest($db, $request, $talk->event_id);
+
+        // edit the talk
+        $talk_mapper->editTalk($data, $talk_id);
+
+        header("Location: " . $request->base . $request->path_info, null, 204);
+        exit;
```
heiglandreas: In my eyes this should return a 200 along with the new talk-data as response.

src/controllers/TalksController.php
```diff
@@ -289,6 +367,12 @@ public function createTalkAction(Request $request, PDO $db)
         $tz = new DateTimeZone($event['tz_continent'] . '/' . $event['tz_place']);
         $talk['date'] = (new DateTime($start_date, $tz))->format('U');
 
+        $event_start_date = (new DateTime($event['start_date']))->format('U');
```
heiglandreas: This uses the servers timezone. Is that wise? Or shouldn't we include the timezone-information here as well?

src/models/TalkMapper.php
```diff
@@ -353,10 +353,79 @@ public function createTalk($data)
...
+     * - description
+     * - slides_link
+     * - language (a value from the column lang:lang_name
+     * - date (a timestamp)
+     * - duration
+     * - speakers (an array of names)
+     * - type_id (id of the talk's type)
+     *
+     * @param $data
+     *
```
heiglandreas: can you add a ```return void``` to show the return-doc hasn't been forgotten?

src/models/TalkMapper.php
```diff
@@ -394,6 +463,57 @@ public function setType($talk_id, $type_id)
...
+     * Add any new speakers not on the list
+     * Remove any speakers already attached the talk that are not in this list
+     * Leave the other speakers alone as they may already have claimed their talk
+     *
+     * @note $speakers contains the display name for each speaker
+     *
+     * @param  int   $talk_id
+     * @param  array $speakers
+     */
```
heiglandreas: Also add a ```return void```

heiglandreas:
I'm missing tests :wink:

liam-wiltshire:
@heiglandreas - Not sure what your view is on this, but assuming none of the stylistic things break it (and granted I've not tested it so I don't know if they do or don't), I'd be tempted to merge this anyway, and then someone can offer another PR to tidy it up? Partly because it's quite an old PR, and from what I know I think @akrabat is pretty tied up at the moment and might not get a chance to look at it in a while (I guess that's for him to comment on), and also because the lack of talk/track admin tools on web2 is the thing we are getting the most complaints about.

That's not to say it shouldn't be tidied up at some point, and hopefully someone will come forward with a PR to do that, but equally if it works, I'm not sure we should potentially lose another 4 months until it gets sorted?

Would appreciate your thoughts :-)

heiglandreas:
@liam-wiltshire I wouldn't wait for tests to merge it! **Especially** as it's a functionality that's been worked on for a long time (I did two approaches to it a year or so also :wink:) and that should be online ASAP! So I'm absolutely with you there!

Please give it a local test to see whether it works as espected and then from my side: merge it!

Perhaps add a new ticket to the issue-tracker to remind us that there are some tests missing :wink:

src/controllers/TalksController.php
akrabat: This follows the convention set for editing an event.

src/controllers/TalksController.php
akrabat: This is just moving existing code to a function for use by both add and edit talk. Original code is here: https://github.com/joindin/joindin-api/blob/master/src/controllers/TalksController.php#L390

i.e. separate ticket to determine if this line is correct or not.

src/models/TalkMapper.php
akrabat: Done.

akrabat:
@liam-wiltshire Updated the items that @heiglandreas raised that I agreed with :)

Outstanding items commented on.

src/controllers/TalksController.php
heiglandreas: I've just had a deeper look at this. It looks like the Server-Timezone is irrelevant, as the date-fields contain the offset so the Server-Timezone is ignored on DateTime-object creation.

So no ticket needed.

heiglandreas:
Fine from my POV

heiglandreas:
Whooo!!!

PR #328, opened 2016-03-27
Add and remove a talk from a track

akrabat:
* Add a track to a talk by POSTing to /talks/123/tracks with the `track_uri` in the body. Publish the `tracks_uri` URI in the talk model representation.
* Remove a track from a talk by DELETEing to /talks/123/tracks/456. Publish the URI to send the DELETE to as `remove_track_uri` in the tracks array on the talk representation, so that it can be discovered.

Note that this PR builds on #321 (Edit track) as it uses the getEventByTrackId() method from that PR.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -188,6 +188,58 @@ class TalksController extends ApiController
     }
 
     /**
+     * Add a track to a talk by POSTing to /talks/123/tracks with the `track_uri`
+     * in the body
+     *
+     * @param PDO $db
+     * @param Request $request
+     */
+    public function addTrackToTalk(Request $request, PDO $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to create data", 400);
+        }
+
+        $talk_id = $this->getItemId($request);
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
+        if (!($is_admin || $is_speaker)) {
+            throw new Exception("You do not have permission to add this talk to a track", 400);
+        }
+
+        $track_uri = $request->getParameter("track_uri");
+        $pattern ='@/' . $request->version . '/tracks/([\d]+)$@';
+        if (!preg_match($pattern, $track_uri, $matches)) {
+            throw new Exception('Invalid track_uri', 400);
+        }
+        $track_id = $matches[1];
+
+        // is this track on this event?
+        $event_mapper = new EventMapper($db, $request);
+        $track_events = $event_mapper->getEventByTrackId($track_id, true, false, false);
+        if (!$track_events || !$track_events[0]['ID']) {
+            throw new Exception("Associated event not found", 400);
+        }
+        $track_event_id = $track_events[0]['ID'];
+        if ($talk->event_id != $track_event_id) {
+            throw new Exception("This talk cannot be added to this track", 400);
+        }
+
+        // add talk to track
+        $talk_mapper->addTalkToTrack($talk_id, $track_id);
+
+        $uri = $request->base . '/' . $request->version . '/talks/' . $talk_id;
+        header('Location: ' . $uri, null, 201);
+        exit;
+    }
+
+    /**
      * Get a single talk
      *
      * @param  PDO      $db
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -240,6 +240,52 @@ class TalksController extends ApiController
     }
 
     /**
+     * Remove a track from a talk by DELETEing to /talks/123/tracks/456
+     *
+     * @param PDO $db
+     * @param Request $request
+     */
+    public function removeTrackFromTalk(Request $request, PDO $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to create data", 400);
+        }
+
+        $talk_id = $this->getItemId($request);
+        $track_id = $request->url_elements[5];
+
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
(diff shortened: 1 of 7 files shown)


Jean85:
This LGTM. I see some duplicated code here and there, but I will gladly refactor that myself afterwards.

This PR is blocking a few others in cascade, through joindin/joindin-web2#338 ... Can we merge this?

heiglandreas:
I'm stumbling over some of the wording. One time it's "*Add a talk to a track*" but most of the time it's "*Add a track to a talk*". Which way around is it? 

For me personally "Add a talk to a track" makes more sense, but that might mean changing the complete API as it would mean calling the "tracks/abc/talks"-endpoint and not "talks/abc/tracks"…

Ideas?

src/controllers/TalksController.php
```diff
@@ -188,6 +188,104 @@ public function deleteAction($request, $db)
...
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
+        if (!($is_admin || $is_speaker)) {
+            throw new Exception("You do not have permission to add this talk to a track", 400);
```
heiglandreas: Can we have a 403 here? Or have a 400 in TracksController:43 :wink:

src/controllers/TalksController.php
```diff
@@ -188,6 +188,104 @@ public function deleteAction($request, $db)
...
+        }
+
+        $track_uri = $request->getParameter("track_uri");
+        $pattern ='@/' . $request->version . '/tracks/([\d]+)$@';
+        if (!preg_match($pattern, $track_uri, $matches)) {
+            throw new Exception('Invalid track_uri', 400);
```
heiglandreas: This should result in a 404, 'cause we can't find valid data on that URI

src/controllers/TalksController.php
```diff
@@ -188,6 +188,104 @@ public function deleteAction($request, $db)
...
+        }
+        $track_id = $matches[1];
+
+        // is this track on this event?
+        $event_mapper = new EventMapper($db, $request);
+        $track_events = $event_mapper->getEventByTrackId($track_id, true, false, false);
+        if (!$track_events || !$track_events[0]['ID']) {
+            throw new Exception("Associated event not found", 400);
```
heiglandreas: Also a 404

src/controllers/TalksController.php
```diff
@@ -188,6 +188,104 @@ public function deleteAction($request, $db)
...
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
+        if (!($is_admin || $is_speaker)) {
+            throw new Exception("You do not have permission to remove this talk from this track", 400);
```
heiglandreas: This should also be a 403 then

src/controllers/TalksController.php
```diff
@@ -188,6 +188,104 @@ public function deleteAction($request, $db)
...
+        }
+
+        // is this track on this event?
+        $event_mapper = new EventMapper($db, $request);
+        $track_events = $event_mapper->getEventByTrackId($track_id, true, false, false);
+        if (!$track_events || !$track_events[0]['ID']) {
+            throw new Exception("Associated event not found", 400);
```
heiglandreas: This should be 404

src/controllers/TracksController.php
```diff
@@ -22,4 +22,79 @@ public function getAction($request, $db)
 
         return $list;
     }
+
+    public function editTrack($request, $db)
+    {
```
heiglandreas: Can this be called without being authorized? I'm missing the authorization check...

src/controllers/TracksController.php
```diff
@@ -22,4 +22,79 @@ public function getAction($request, $db)
...
+        }
+        if ($errors) {
+            throw new Exception(implode(". ", $errors), 400);
+        }
+
+        $track_mapper->editEventTrack($track, $track_id);
+
+        $uri = $request->base  . '/' . $request->version . '/tracks/' . $track_id;
+        header("Location: $uri", true, 204);
+        exit;
```
heiglandreas: Shouldn't a POST return the updated resource?

src/controllers/TracksController.php
```diff
@@ -22,4 +22,79 @@ public function getAction($request, $db)
...
+    }
+
+    public function deleteTrack($request, $db)
+    {
```
heiglandreas: Also missing authentication-check…

src/models/EventMapper.php
```diff
@@ -100,6 +100,32 @@ public function getEventById($event_id, $verbose = false, $activeEventsOnly = tr
...
+     * @param bool $activeEventsOnly
+     * @param bool $transform
+     *
+     * @return array the event detail
+     */
+    public function getEventByTrackId($track_id, $verbose = false, $activeEventsOnly = true, $transform = true)
+    {
+        $results = $this->getEvents(1, 0, array("track_id" => $track_id, 'active' => $activeEventsOnly));
+        if (!$results) {
+            return false;
```
heiglandreas: Wouldn't it be easier to throw an exception here? Now we always have to check whether ```getEventByTrackId``` returns ```false``` to take appropriate measures. With an Exception that would be wrapping the call into a ```try-catch```-block or just passing the exception on…

src/models/TalkMapper.php
```diff
@@ -589,6 +591,56 @@ public function thisUserHasAdminOn($talk_id)
...
+        return $talk_track_id;
+    }
+
+    /**
+     * Remove talk from a track
+     *
+     * @param  int $talk_id
+     * @param  int $track_id
+     */
+    public function removeTrackFromTalk($talk_id, $track_id)
```
heiglandreas: Either we're adding a talk to a track (then we also need to remove that talk from the track) or we add a track to a talk (then we also need to remove the track from the talk). Which way around?

heiglandreas:
I'm missing tests :wink:

PR #330, opened 2016-04-20
Add link to reported comments page

kennyray:
I added a function which will wrap the passed in event name with a link to the reported comments page, and changed the code that was simply displaying the name of the event so that it calls that function.

### DIFF
--- a/src/services/EventCommentReportedEmailService.php
+++ b/src/services/EventCommentReportedEmailService.php
@@ -59,7 +59,8 @@ class EventCommentReportedEmailService extends EmailBaseService
 
     private function linkToReportedCommentsForEvent($name)
     {
-        return '<a href="' . $this->website_url . '/event/' . strtolower($name) 
+          return '<a href="' . $this->website_url
+                . '/event/' . strtolower($name)
                 . '/reported-comments' . '">' . $this->event['name'] . '</a>';
     }
 }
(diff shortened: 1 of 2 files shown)


lornajane:
ok to test

akrabat:
Please do this in Markdown within eventCommentReported.md so that the plain text version of the email doesn't end up with HTML in it.

liam-wiltshire:
@kennyray thanks for your contribution.

Couple of issues however - the returned URL doesn't appear to be correct - I've just done a test on an event with the slug 'it-day', but in the email the slug is 'it day' (with a space). 

If you have a look in the EventMapper, there is a field 'url_friendly_name' which contains the sanitised slug - if you can use this instead of `strtolower($event_name)` I think that should sort it.

Appreciate you posted this PR a while ago (sorry!!), so if you haven't got time to look at it right now, let me know and I'll see if I can sort it.

Thanks,

Liam

kennyray:
Liam,

Sorry that I didn't get that right. My job responsibilities have shifted and I'm focused on C# now. Haven't been able to touch PHP for a while, so if you could make that adjustment for me I'd appreciate it. :-)

Kenny

liam-wiltshire:
Opened new PR merging @kennyray's commits with additional fixes - https://github.com/joindin/joindin-api/pull/343

PR #331, opened 2016-05-07
Error if event start date is later than its end date

jonnixs:


src/controllers/EventsController.php
```diff
@@ -203,6 +203,8 @@ public function postAction($request, $db)
             $end_date   = strtotime($request->getParameter("end_date"));
             if (! $start_date || ! $end_date) {
                 $errors[] = "Both 'start_date' and 'end_date' must be supplied in a recognised format";
+            } else if ( $start_date > $end_date ) {
```
johnkary: Hi @phoenixrises! The Jenkins build shows a build failure because these changes don't match the project's coding style. This is a super easy fix by changing this line (and the second instance of it in the same file below) to instead use this style:

    } elseif ($start_date > $end_date) {

Could you push a last commit with this update to help the build pass? Then it might be more likely to be merged by the project maintainers.

Thanks!
🚀

phoenixrises: @johnkary Fixed. Ta.

Jean85:
I would add a test that checks for no errors when an event starts and ends on the same day.

Other than that, LGTM (tested locally).

PR #332, opened 2016-05-16
Fix for average rating persists when rating is removed

leightonthomas:
Jira: https://joindin.jira.com/browse/JOINDIN-687

This adds an extra check in the `get_talk_rating` function to see if the comment is marked as active (not 0). I'm not aware of any instances other than reporting comments where a comment would be marked as inactive, but if there are any then let me know and I'll make changes :)

### DIFF
--- a/db/patch66.sql
+++ b/db/patch66.sql
@@ -31,3 +31,5 @@ BEGIN
 
 	RETURN rating_out;
 END;
+
+INSERT INTO patch_history SET patch_number = 66;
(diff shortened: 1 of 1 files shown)


Jean85:
Duplicated of #335

liam-wiltshire:
@leightonthomas thanks for your PR, unfortunately it was duplicate in another PR which has been merged - identical changes, so thanks for essentially verifying that the commit worked!!

Please do submit more PRs - see https://joindin.jira.com/issues/?filter=10510 for a few ideas!

PR #335, opened 2016-07-03
Average rating persists when rating is removed

RobWillison:
In relation to [JOINDIN-687](https://joindin.jira.com/browse/JOINDIN-687)

The get_talk_rating mysql function now checks the active flag in the talk_comments so
the average comment rating doesn't include removed/reported comments.

### DIFF
--- a/db/patch66.sql
+++ b/db/patch66.sql
@@ -1,0 +1,35 @@
+-- Drop the function if it exists
+DROP FUNCTION IF EXISTS get_talk_rating;
+
+
+-- Create get_talk_rating function that takes into account ratings that should not be added (speaker ratings, private)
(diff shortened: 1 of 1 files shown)


lornajane:
ok to test

liam-wiltshire:
Hah, think I'm the third person to write this patch - wasn't claimed on JIRA :-).

But yes, this patch works.

Jean85:
For the record, the only change compared to prev. version of the function (patch 53.sql) is the added condition `tc.active = 1`

LGTM :+1:

liam-wiltshire:
Jean85 - thanks for confirming, going to try and get some of these PRs merged over the next few days.

PR #336, opened 2016-07-18
Require security patched version of guzzle

AndrewCarterUK:
Fixes: https://httpoxy.org/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,7 @@
     "require" : {
         "swiftmailer/swiftmailer": "v5.0.3",
         "michelf/php-markdown": "1.4.0",
+        "guzzlehttp/guzzle": "4.2.4",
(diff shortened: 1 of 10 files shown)


akrabat:
ok to test

composer.json
```diff
@@ -3,6 +3,7 @@
     "require" : {
         "swiftmailer/swiftmailer": "v5.0.3",
         "michelf/php-markdown": "1.4.0",
+        "guzzlehttp/guzzle": "4.2.4",
```
akrabat: Don't we want `~4.2.4` here?

AndrewCarterUK: Happy to change it, should I change the others too?

akrabat: We'll fix the others separately. Someone should go through and check if there's been any security updates to those libraries too…

akrabat:
Thanks!

PR #337, opened 2016-08-18
Fix path for event icons

Jean85:
As in title. Discovered this with @akrabat in IRC..

### DIFF
--- a/src/config.php.dist
+++ b/src/config.php.dist
@@ -13,7 +13,7 @@ $config =  array(
             'web2',
         ),
     ),
-    'event_image_path' => __DIR__ . '/../../joind.in/src/inc/img/event_icons/',
+    'event_image_path' => __DIR__ . '/../../joindin-legacy/src/inc/img/event_icons/',
     'email' => array(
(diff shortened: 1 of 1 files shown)


akrabat:
ok to test

Jean85:
This isn't testable without https://github.com/joindin/joindin-vm/pull/89
Without that you can't upload anything locally. Also, I think this needs some further work, because after a successful upload the link given by the API seems wrong. 

Right now I get `https://joind.in/inc/img/event_icons/icon-4-small.jpg` which is wrong multiple times:
 * no `dev.` in base URL
 * even using `dev.` manually, I get a `Refused connection` error (maybe Apache is misconfigured?)

Jean85:
Ok now it seems to work, tested locally.

liam-wiltshire:
Seems to work for me too

Jean85:
Can we merge this then?

heiglandreas:
It's not critical in that it will bring the live-servers down as it's the fallback for local installations. You tested it localy and it worked. @liam-wiltshire Tested it as well. So **I'd** merge it!

PR #338, opened 2016-08-21
Ability to delete a user (satisfies JOINDIN-559)

liam-wiltshire:
The API side of the PR in web2 - creates a DELTE verb for the user, and adds a delete method to the user mapper.


src/config/routes/2.1.json
```diff
@@ -144,6 +144,13 @@
     },
 
     {
+        "path": "/users(/[^/]+)*/?$",
+        "controller": "UsersController",
+        "action": "deleteAction",
```
heiglandreas: I would name the action ```deleteUser```

src/controllers/UsersController.php
```diff
@@ -298,4 +298,36 @@ public function passwordReset(Request $request, $db)
...
+
+
+    public function deleteAction($request, $db)
+    {
+
+
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
+        }
+        // delete the talk
```
heiglandreas: Please either remove or alter the comment ;)

src/config/routes/2.1.json
liam-wiltshire: @heiglandreas Within the API controllers, they all follow the format [verb]Action  (getAction, postAction etc), so I believe this to be correct.

src/controllers/UsersController.php
```diff
@@ -298,4 +298,36 @@ public function passwordReset(Request $request, $db)
...
+        $user_id     = $this->getItemId($request);
+        $user_mapper = new UserMapper($db, $request);
+
+        // note: use the mapper's getUserById as we don't want to throw a not found exception
+        $user = $user_mapper->getUserById($user_id);
+        if (false === $user) {
+            // user isn't there so it's as good as deleted
+            header("Content-Length: 0", null, 204);
```
heiglandreas: Isn't that more a 404? The resource wasn't found and we couldn't fulfill the request?

liam-wiltshire: All the response codes etc are based on what's already there in the other APIs (I think specifically this was based on the delete method for deleting a starred talk)

src/controllers/UsersController.php
```diff
@@ -298,4 +298,36 @@ public function passwordReset(Request $request, $db)
...
+            exit; // no more content
+        }
+
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
```
heiglandreas: Shouldn't that be a 403?

src/config/routes/2.1.json
heiglandreas: Sorry to prove you wrong, but have a look a few lines down... ```action: updateUser```

And I had a discussion some time ago with Lorna and Rob and at least at that time we agreed on making the actions more "speaking" (at least that's what I remember)

src/controllers/UsersController.php
liam-wiltshire: Take your point, however I've based it off of the deleteAction in https://github.com/joindin/joindin-api/blob/master/src/controllers/TalksController.php - so I've gone for consistency, the same errors, the same HTTP codes etc

src/controllers/UsersController.php
heiglandreas: This is a somewhat philosophical thing. In my eyes (and I seem not to be the only person - http://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete#comment-10798415) a call to delete a non-existing resource is not a valid request.

Besides that, there should be room for improvement. But I'm with you, that we should be consistent and even if that's consistently strange :wink:

src/models/UserMapper.php
```diff
@@ -642,4 +642,14 @@ public function resetPassword($token, $password)
...
     }
+
+
+    public function delete($user_id)
+    {
+        $sql  = "delete from user where ID = :user_id";
+        $stmt = $this->_db->prepare($sql);
+        $stmt->execute(array("user_id" => $user_id));
+
+        return true;
```
heiglandreas: How can this always return ```true```? Shouldn't we check the result of the transaction and decide on that what to return?

src/controllers/UsersController.php
```diff
@@ -298,4 +298,36 @@ public function passwordReset(Request $request, $db)
...
+        }
+
+        $user_mapper->delete($user_id);
```
heiglandreas: If we check the result of the delete-transaction we should here decide whether to return a 400 or 204

src/config/routes/2.1.json
liam-wiltshire: I'm not precious about it either way - most of the code was a direct copy from another deleteAction if I'm honest - will change the method name and commit.

src/models/UserMapper.php
liam-wiltshire: Probably :-), again I was just basing it on the delete method in https://github.com/joindin/joindin-api/blob/master/src/models/TalkMapper.php but I'm happy to change it.

liam-wiltshire:
@heiglandreas - made some amends based on your feedback, it is cleaner like that - guess it serves me right for trying to shortcut and copying old code :-)

src/controllers/UsersController.php
```diff
@@ -298,4 +298,38 @@ public function passwordReset(Request $request, $db)
...
+
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
+        }
+        
+        if ($user_mapper->delete($user_id)){
+            header("Content-Length: 0", null, 204);
+            exit; // no more content            
+        }else{
```
heiglandreas: You can omit the ```else``` here. 

When the ```if``` was used, the script ended with the ```exit``` :wink:

src/controllers/UsersController.php
```diff
@@ -298,4 +298,38 @@ public function passwordReset(Request $request, $db)
         }
 
     }
+
```
heiglandreas: There is a lot of empty lines. Are they really necessary? :wink:

liam-wiltshire: Well, my house is very small, so putting extra space into code makes me feel better :-p

Will edit and push

src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
...
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
+        }
+        
+        if ($user_mapper->delete($user_id)){
+            header("Content-Length: 0", null, 204);
+            exit; // no more content            
+        }
+        
+        throw new Exception("There was a problem trying to delete the user", 400);
```
heiglandreas: One more thing:

I think it's cleaner to have the "default" stuff as the main path and error-conditions "branching off".

That would mean in this case, throw the Exception within the ```if``` and have the header and the exit as last statements in the function.

src/controllers/UsersController.php
Jean85: I don't think it's strange. It makes the DELETE action idempotent, as it should be. If, for example, a mobile app retries the DELETE action because it missed the response on the previous request, it won't show an error.

heiglandreas: Wether an app shows an error should not be the reason which response an API gives. And when a DELETE is tried multiple times the action can be sucessful the first time and after that you are trying to delete a non-existing resource which should give a 404.

An app that is using that API should know that after deleting a resource a 404 is about as good as a 204 as the resource isn't there (anymore). Whether that's the result of the own action or due to someone else deleting the resource in between shouldn't matter. 

My Personal POV on a philosophical (and much debated) question :wink:

liam-wiltshire:
Closing to fix my screwed up branches - will re-open.

PR #339, opened 2016-08-26
Allow 401-Headers 

heiglandreas:
This PR allows 401-Headers for actions that require authentication. 

These actions are currently returning a 400-Status which I'd consider too generic.

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -149,7 +149,7 @@ class EventsController extends ApiController
     public function postAction($request, $db)
     {
         if (! isset($request->user_id)) {
-            throw new Exception("You must be logged in to create data", 400);
+            throw new Exception("You must be logged in to create data", 401);
         }
         if (isset($request->url_elements[4])) {
             switch ($request->url_elements[4]) {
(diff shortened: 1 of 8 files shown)


src/controllers/EventsController.php
```diff
@@ -539,6 +539,7 @@ function ($tag) {
      */
     public function createTrack($request, $db)
     {
+        // Should we not check for a login?
```
heiglandreas: And fail fast

src/controllers/TracksController.php
```diff
@@ -25,6 +25,7 @@ public function getAction($request, $db)
 
     public function editTrack($request, $db)
     {
+        // Should we not check for login?
```
heiglandreas: And fail fast

src/controllers/TracksController.php
```diff
@@ -74,6 +75,7 @@ public function editTrack($request, $db)
 
     public function deleteTrack($request, $db)
     {
+        // Should we not check for login?
```
heiglandreas: And fail fast

Jean85:
I agree with the fail-fast approach! LGTM! :+1:

liam-wiltshire:
@heiglandreas yes, we should be checking for login on tracks - it's not critical in the sense of it being a security hole, as we are checking that the user can edit that event, but equally we could advise some unnecessary  DB reads by doing so.

Will raise a ticket in JIRA to add that in - would be a good one for the hackathon.

Just testing the other changes now, but I can't see any issues so far, so all being well will merge later this eve.

PR #341, opened 2016-09-02
Hotfix/fixtests

heiglandreas:
This PR fixes the current testsuite by adapting the calls to phpunits deprecated ```getMock```-methods to either ```createMock``` or to ```getMockObject```

During this process I stumbled over two tests created by @choult that are either obsolete by now or that were working already but have been marked to skip. Can you shed some light there please?

### DIFF
--- a/src/inc/Request.php
+++ b/src/inc/Request.php
@@ -134,7 +134,7 @@ class Request
             $header = new Header('X-Forwarded-For', $_SERVER['HTTP_X_FORWARDED_FOR'], ',');
             $header->parseParams();
             $elementArray = $header->buildEntityArray();
-            $ipAddress    = count($elementArray) ? $elementArray[0] : null;
+            $ipAddress    = count($elementArray) ? $elementArray[0][0] : null;
         }
         $this->clientIP        = $ipAddress;
         $this->clientUserAgent = $userAgent;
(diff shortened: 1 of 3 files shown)


choult:
@heiglandreas Oo, that's a bit old... I suspect that was a bit naughty - found a problem via testing and then didn't fix it... There might be a corresponding ticket in JIRA somewhere

choult:
As to the VersionedRouter test, yes we still need it (we're using VersionedRouter in the front controller) but you can probably convert the mocked VersionedRouter to a concrete one and remove the reference to `getLegacyRoute` - it looks like that was a piece of scaffolding that is no longer required.

heiglandreas:
I didn't want to remove all of the VersionedRouterTests :wink: I'll have a look at the ```getRoute```as that one seemed to expect that legacy method...

Thanks for the insight @choult!

liam-wiltshire:
The changes to replace getMockBuilder etc are all good, but I'm a little nervous about removing the Timezone class - I can't see why it would have been written in the first place if it wasn't needed...

That said, perhaps it is just something that has been 'written out' of the system over time and is no-longer required, so perhaps it just needs a close going over to ensure it doesn't have any knock-on elsewhere?

heiglandreas:
I checked for occurrences of calls to the Timezone-class and couldn't find some apart from the removed ones. But you are right, it'S out of the scope of the test-fixes so I should move that to a separate PR with separate tests.

liam-wiltshire:
OK cool, I'll leave that with you then :-)

heiglandreas:
Moved the Timezone-stuff into https://github.com/joindin/joindin-api/pull/344

liam-wiltshire:
Perfect- looks good, but I'm on my mobile, will merge when I get a chance.

PR #342, opened 2016-09-09
API route, controller and mapper method to delete user

liam-wiltshire:
Required for https://github.com/joindin/joindin-web2/pull/349

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -298,4 +298,26 @@ class UsersController extends ApiController
         }
 
     }
+
+    public function deleteUser($request, $db)
+    {
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
+        }
+        // delete the user
+        $user_id     = $this->getItemId($request);
+        $user_mapper = new UserMapper($db, $request);
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
+        }
+
+        if (! $user_mapper->delete($user_id)){
+            throw new Exception("There was a problem trying to delete the user", 400);
+        }
+        header("Content-Length: 0", null, 204);
+        exit; // no more content
+
+    }
+
 }
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -299,14 +299,19 @@ class UsersController extends ApiController
 
     }
 
-    public function deleteUser($request, $db)
+    public function deleteUser($request, $db, $user_mapper = false)
     {
         if (! isset($request->user_id)) {
             throw new Exception("You must be logged in to delete data", 400);
         }
         // delete the user
         $user_id     = $this->getItemId($request);
-        $user_mapper = new UserMapper($db, $request);
+
+        //Add ability to pass $user_mapper in, to allow for a unit test,
+        //Without breaking backwards compatibility
+        if (!$user_mapper) {
+            $user_mapper = new UserMapper($db, $request);
+        }
         $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
         if (! $is_admin) {
             throw new Exception("You do not have permission to do that", 400);
@@ -315,6 +320,9 @@ class UsersController extends ApiController
         if (! $user_mapper->delete($user_id)){
             throw new Exception("There was a problem trying to delete the user", 400);
         }
+        //If we are unit testing, then we can't exit or send headers!
+        if (defined('UNIT_TEST')) return true;
+
         header("Content-Length: 0", null, 204);
         exit; // no more content
 
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -302,7 +302,7 @@ class UsersController extends ApiController
     public function deleteUser($request, $db, $user_mapper = false)
     {
         if (! isset($request->user_id)) {
-            throw new Exception("You must be logged in to delete data", 400);
+            throw new Exception("You must be logged in to delete data", 401);
         }
         // delete the user
         $user_id     = $this->getItemId($request);
@@ -314,7 +314,7 @@ class UsersController extends ApiController
         }
         $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
         if (! $is_admin) {
-            throw new Exception("You do not have permission to do that", 400);
+            throw new Exception("You do not have permission to do that", 403);
         }
 
         if (! $user_mapper->delete($user_id)){
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -309,7 +309,7 @@ class UsersController extends ApiController
             throw new Exception("You must be logged in to delete data", 401);
         }
         // delete the user
-        $user_id     = $this->getItemId($request);
+        $user_id = $this->getItemId($request);
 
         $user_mapper = $this->getUserMapper($db, $request);
 
@@ -318,27 +318,30 @@ class UsersController extends ApiController
             throw new Exception("You do not have permission to do that", 403);
         }
 
-        if (! $user_mapper->delete($user_id)){
+        if (! $user_mapper->delete($user_id)) {
             throw new Exception("There was a problem trying to delete the user", 400);
         }
         //If we are unit testing, then we can't exit or send headers!
-        if (defined('UNIT_TEST')) return true;
+        if (defined('UNIT_TEST')) {
+            return true;
+        }
 
         header("Content-Length: 0", null, 204);
         exit; // no more content
 
     }
 
-    public function setUserMapper(UserMapper $user_mapper){
+    public function setUserMapper(UserMapper $user_mapper)
+    {
         $this->user_mapper = $user_mapper;
     }
 
-    public function getUserMapper($db, $request){
-        if (! $this->user_mapper){
+    public function getUserMapper($db, $request)
+    {
+        if (! $this->user_mapper) {
             $this->user_mapper = new UserMapper($db, $request);
         }
 
         return $this->user_mapper;
     }
-
 }
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -302,6 +302,7 @@ class UsersController extends ApiController
 
     }
 
+
     public function deleteUser($request, $db)
     {
         if (! isset($request->user_id)) {
@@ -311,6 +312,7 @@ class UsersController extends ApiController
         $user_id     = $this->getItemId($request);
 
         $user_mapper = $this->getUserMapper($db, $request);
+
         $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
         if (! $is_admin) {
             throw new Exception("You do not have permission to do that", 403);
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -2,6 +2,9 @@
 
 class UsersController extends ApiController
 {
+
+    protected $user_mapper;
+
     public function getAction($request, $db)
     {
         $user_id = $this->getItemId($request);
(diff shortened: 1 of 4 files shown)


src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
...
+        $user_mapper = new UserMapper($db, $request);
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
+        }
+
+        if (! $user_mapper->delete($user_id)){
+            throw new Exception("There was a problem trying to delete the user", 400);
+        }
+        header("Content-Length: 0", null, 204);
```
heiglandreas: Should we not add a ```Location:```-Header to the user-list (or the main site) so that a client knows where to go next?

src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
...
+    {
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
+        }
+        // delete the user
+        $user_id     = $this->getItemId($request);
+        $user_mapper = new UserMapper($db, $request);
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
```
heiglandreas: In my eyes this should be a ```403 Not Authorized```

src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
         }
 
     }
+
+    public function deleteUser($request, $db)
+    {
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
```
heiglandreas: In my eyes this shoud be a ```401 Unauthorized```

src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
...
+        // delete the user
+        $user_id     = $this->getItemId($request);
+        $user_mapper = new UserMapper($db, $request);
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
+        if (! $is_admin) {
+            throw new Exception("You do not have permission to do that", 400);
+        }
+
+        if (! $user_mapper->delete($user_id)){
+            throw new Exception("There was a problem trying to delete the user", 400);
```
heiglandreas: Shouldn't we differ here between "The resource never existed and therefore couldn't be removed" and "There was an error during execution and the action cound't complete successfully"?

src/controllers/UsersController.php
```diff
@@ -298,4 +298,26 @@ public function passwordReset(Request $request, $db)
...
+
+    public function deleteUser($request, $db)
+    {
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
+        }
+        // delete the user
+        $user_id     = $this->getItemId($request);
+        $user_mapper = new UserMapper($db, $request);
+        $is_admin = $user_mapper->thisUserHasAdminOn($user_id);
```
heiglandreas: What happens when ```thisUserHasAdminOn``` is called with a non existent User-ID?

heiglandreas:
Can we have some unit-tests please? :wink: Especially deleting users should in my eyes be automatically tested to be sure that it works as expected...

liam-wiltshire:
Added tests, had to make a few modifications to the method to facilitate/work around a couple of things.

Appreciate some of your other points @heiglandreas, however in my opinion I think we should aim for consistency with return codes etc - if then there is a scope to update *all* the APIs to new values (but still be consistent), then that could/should be handled as a separate thing- what do you think?

heiglandreas:
The return-types are already inconsistent 😦. Have a look f.e. at https://github.com/joindin/joindin-api/blob/master/src/controllers/UsersController.php#L49 or https://github.com/joindin/joindin-api/blob/master/src/controllers/UsersController.php#L56 where we return a 404 when the resource isn't found. 

So the Return-Types are called used. The same there with the 403 (https://github.com/joindin/joindin-api/blob/master/src/controllers/UsersController.php#L198). 

In my eyes it would make sense to handle that right from the start of a new API-Endpoint not necessarily in a consistent way (as it's consistently inconsistent) but in a way a user of the API might expect it to be used. 

And changing the return types could be seen as a BC-break so the less we have of those - should we change the return code handling - the better.

But I'm not going to stand against that approach to add consistency with return types in separate PRs (Already started https://github.com/joindin/joindin-api/pull/339 😉 )

tests/controllers/UsersControllerTest.php
```diff
@@ -0,0 +1,166 @@
...
+     * Ensures that if the deleteUser method is called and no user_id is set,
+     * an exception is thrown
+     *
+     * @return void
+     *
+     * @test
+     * @expectedException \Exception
+     * @expectedExceptionMessage You must be logged in to delete data
+     */
+    public function testDeleteUserWithNoUserIdThrowsException(){
```
heiglandreas: From what I've seen so far the Codestyling here is different from the styling used in the ```src```-folder. Is that on purpose? The ```tests``` -folder is currently not checked by phpcs as I've noticed.

liam-wiltshire: There was a very good reason for this - I was being lazy :-).

Have sorted, will look at those return codes, and as long as they don't break anything (I don't *think* they will, but will test), then I'll update them.

liam-wiltshire:
Doesn't seem to cause any issues, so I've changed those response codes.

As such, from my side, I think it's ready to merge.

heiglandreas:
Just one logical question after the technical side seems to be resolved: What happens with content that's associated with the user? Should that be removed as well? If not how do we handle removed users with such content (ratings f.e.)?

akrabat:
I think that in the past, the comments have been set to anonymous. @lornajane should be able to confirm, though.

liam-wiltshire:
Yes, exactly that - any talks they have claimed go back to showing an unlinked name, and any comments they've posted show as "Anonymous"

heiglandreas:
I've just done a little research. 

```v2.1/talks/[talk-id]/comments```: With the current method implemented the ```user-uri``` of a comment is still returning the (now missing) URI to the deleted user, ```user_display_name``` is ```null``` though.

```v2.1/events/[event-id]?verbose=true```: User is not displayed as host

```v2.1/events/event-id]/attendees```: User is not displayed, though the attendee-count is different from ```v2.1/events/[event-id]```

So from my POV the ```DELETE``` should also remove the entries in the ```user_attend``` and ```user_admin```-tables as well as set the user-id in the ```talk_comments```-table to ```0```.

What do you think @liam-wiltshire, @akrabat and @lornajane?

**Comment by liam-wiltshire** [comment]
I guess it comes down to a scale between 'getting it working' to 'technical perfection', or indeed somewhere along the line.

As it stands, my personal aims are to a) Get as many of the missing features put onto web2 as possible, and b) Have plenty of 'easypick' and 'hackathon' issues in JIRA to tempt people on board - as such, as much as I know the 'getting it working' end of the scale will result in technical debt, it also probably gives an opportunity for an 'easypick' item.

As such, currently I lean toward the view that as long as it works (and having tested web2 I believe it does), then ship it so that the community can see we are putting these required features in, raise an issue in JIRA as an improvement, and then look for someone to pick it up as an introduction to Joind.In contribution?

That's just my view however - others may see it differently! :-)

**Comment by heiglandreas** [comment]
That's a perfect approach in my eyes. It'll get it shipped ASAP and we have some easy-pick issues for the hackaton. I think adding the appropriate-delete-entries to the current implementation does qualify as easy-pick. If we go that way I'll add the issues I raised to jira later today and mark them easy pick (when I'm allowed to do so :wink:)

File: src/controllers/UsersController.php [hunk 7]
```diff
@@ -298,4 +298,34 @@ public function passwordReset(Request $request, $db)
         }
 
     }
+
+    public function deleteUser($request, $db, $user_mapper = false)
```
akrabat: Do not pass in `$user_mapper` to this method as it's a dependency on the class. Create a `setUserMapper()`/`getUserMapper()` and have `getUserMapper()` instantiate if `$this->user_mapper` hasn't already been set.

liam-wiltshire:
Changes made to move $user_mapper outside of the method - probably needs applying to the other methods for consistency, but that can be picked up at a later point.

liam-wiltshire:
ok to test

tests/controllers/UsersControllerTest.php
```diff
@@ -0,0 +1,161 @@
...
+     * @test
+     * @expectedException        \Exception
+     * @expectedExceptionMessage You must be logged in to delete data
+     */
+    public function testDeleteUserWithNoUserIdThrowsException()
+    {
+        $request = new \Request([], ['REQUEST_URI' => "http://api.dev.joind.in/v2.1/users/3", 'REQUEST_METHOD' => 'DELETE']);
+
+        $usersController = new \UsersController();
+        $db = $this->getMockBuilder('\JoindinTest\Inc\mockPDO')->getMock();
```
heiglandreas: You could also use ```$this->getMockBuilder('PDO')->disableOriginalConstructor()->getMock();``` to Mock the PDO-Instance. Pro *and* Con is you need to mock the called methods as well. :wink:

heiglandreas: YOu might want to have a look at https://github.com/joindin/joindin-api/pull/340/files#diff-2e9dde7e960c27b3feffd135af2229adR108

src/controllers/UsersController.php
```diff
@@ -298,4 +301,47 @@ public function passwordReset(Request $request, $db)
...
     }
+
+
+    public function deleteUser($request, $db)
+    {
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 401);
+        }
+        // delete the user
+        $user_id     = $this->getItemId($request);
```
akrabat: Remove additional whitespace here - no need to line up as it's on its own

src/models/UserMapper.php
```diff
@@ -642,4 +642,15 @@ public function resetPassword($token, $password)
 
         return false;
     }
+
+    public function delete($user_id)
+    {
+        try {
+            $sql = "delete from user where ID = :user_id";
+            $stmt = $this->_db->prepare($sql);
+            return $stmt->execute(array("user_id" => $user_id));
```
akrabat: Shouldn't we also be unsetting this user id from all the comments/attended by tables?

liam-wiltshire: @heiglandreas and I discussed this about 14 days ago - as it stands, the API and Web2 handle it between them in that just by deleting the user, talks that were linked to that user become 'unlinked' and comments become 'anonymous' in the frontend by virtue of the user not existing.

As such, we suggested it worked as it was, but that would make a potential 'Easy Pick' or 'Hackathon' issue.

Having said that, if you'd rather it got sorted now, I can look at it.

akrabat: Sorry for the slow reply.

I think it should be done now as the user id should be purged.

src/models/UserMapper.php
```diff
@@ -642,4 +642,43 @@ public function resetPassword($token, $password)
 
         return false;
     }
+
+    public function delete($user_id)
+    {
+        try {
```
heiglandreas: Shouldn't we start a transaction before doing all those deletions? As currently the state will be somewhat undefined when an exception is thrown as we don't know at what point that exception was thrown. With a transaction the last point within the try would be a commit and within the catch we'd do a rollback and be in a known state (no user deleted)

railto: Nice catch @heiglandreas, that should be added for sure

PR #344, opened 2016-09-13
Removes Timezone-class

heiglandreas:
This commit removes the Timezone-class that was used at two locations by altering those two locations to a more recent approach by adding calls to the DateTime-family directly in the code.

This also removes the test-file.


heiglandreas:
There you go ;)

liam-wiltshire:
Thanks, will give it a once-over and merge this evening.

liam-wiltshire:
@heiglandreas - I've tried this locally and it doesn't seem to work.

If I checkout the master branch, http://api.dev.joind.in/v2.1/events returns events as normal, however if I checkout your branch, I get:

0: DateTime::__construct(): Failed to parse time string (1536695254) at position 8 (5): Unexpected character

Does this not happen on your local?

Liam

heiglandreas:
Thanks for noting. I have no Idea *what* I tested during creation but it seemed not to be the change i introduced here.... 

Fixed the issue though!

liam-wiltshire:
This seems ok now, thanks.

PR #345, opened 2016-09-22
Allow event hosts to edit talks

ihabunek:
Hi! I believe this should work but was unable to test because the setup procedure described in `README-local-dev.md` did not work for me. :/

### DIFF
--- a/src/models/TalkMapper.php
+++ b/src/models/TalkMapper.php
@@ -72,7 +72,7 @@ class TalkMapper extends ApiMapper
             $results = $stmt->fetchAll(PDO::FETCH_ASSOC);
             $total = $this->getTotalCount($sql, array(':event_id' => $event_id));
             $results = $this->processResults($results);
-            
+
             return new TalkModelCollection($results, $total);
         }
 
@@ -697,7 +697,7 @@ class TalkMapper extends ApiMapper
                     . ' from user_admin a '
                     . ' inner join user u on u.ID = a.uid '
                     . ' inner join talks t on t.event_id = rid '
-                    . ' where rtype="event" and rcode!="pending"'
+                    . ' where rtype="event" and (rcode!="pending" OR rcode is null)'
                     . ' AND u.ID = :user_id'
                     . ' AND t.ID = :talk_id';
             $stmt = $this->_db->prepare($sql);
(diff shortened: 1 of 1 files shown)


lornajane:
It would be worth checking but the TalkMapper::thisUserHasAdminOn method _should_ handle the case where the user is an event admin.  This method is re-used in a number of places so this fix should be applied there if possible

ihabunek:
Hm then the question would be why does this SQL code returns no rows in my case...
https://github.com/joindin/joindin-api/blob/master/src/models/TalkMapper.php#L695-L702

It's a bit hard to test since the dev setup doesn't work, and I don't have the time to debug now.

My actual issue on production is that i'm unable to edit talks for this event, and I'm the host:
https://joind.in/event/view/6049

liam-wiltshire:
@ihabunek, @lornajane  I'm going to look into this on test. - it does look like that query should account for event admins, but likewise having read it, I'm not 100% sure if perhaps it's not correct.

Will update the test. database and have a play. however I think that if that query is incorrect, then we are probably better off fixing it in there, rather than editing a single use case?

ihabunek:
Ok, I think I see the difference.

From `thisUserHasAdminOn`:
```sql
where rtype="event" and rcode!="pending"
```

From `thisUserHasAdminOn`:
```sql
rtype="event" and (rcode!="pending" OR rcode is null)
```
https://github.com/joindin/joindin-api/blob/master/src/models/TalkMapper.php#L700
https://github.com/joindin/joindin-api/blob/master/src/models/EventMapper.php#L515

liam-wiltshire:
@ihabunek thanks for flagging this - I've just tested on the test platform and I get the same issue, so that does seem to be the case.

I don't have access to the VM right now so I can't check that adding `is null` is the fix, but it seems the likely candidate.

As such, are you able to back out your above change and commit with the 'new' fix, and I'll check it when I am home this evening?

Thanks!

ihabunek:
@liam-wiltshire There, pushed the new fix. Thanks to you and @lornajane for the review.

BTW, `col != "value"` doesn't match nulls? MySQL sadness. :/

liam-wiltshire:
One of the many joys of mySQL (!)

Thanks @ihabunek, will check it in a VM later and merge.

liam-wiltshire:
ok to test

liam-wiltshire:
ok to test

liam-wiltshire:
ok to test

PR #352, opened 2016-10-04
Post talk speaker endpoint - JOINDIN-639

liam-wiltshire:
This creates the API endpoint required in https://joindin.jira.com/browse/JOINDIN-639

It's complete, except the sending emails (as the emails need a link, which requires the work to be done in web-2 as well), but before I moved on any further, I largely wanted a more in-depth review, as I've not really done any 'big' things on the API side before!

Functionally it all works, but there might be better ways to juggle the mappers etc.

In short, consider this a request for a code review :-)

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -503,7 +503,8 @@ class TalksController extends ApiController
                 $speaker = filter_var($speaker, FILTER_SANITIZE_STRING);
                 $speaker = trim($speaker);
                 return $speaker;
-            }, (array) $request->getParameter('speakers')
+            },
+            (array) $request->getParameter('speakers')
         );
 
         return $talk;
@@ -538,7 +539,7 @@ class TalksController extends ApiController
         $data = $this->getLinkUserDataFromRequest($request);
 
         if ($data['display_name'] == '' || $data['username'] == '') {
-            throw new Exception("You must provide a display name and a username",400);
+            throw new Exception("You must provide a display name and a username", 400);
         }
 
         //Get the speaker record based on the display name - check if this is already claimed,
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -519,7 +519,6 @@ class TalksController extends ApiController
 
     public function setSpeakerForTalk(Request $request, PDO $db)
     {
-        //We must be logged in
         if (!isset($request->user_id)) {
             throw new Exception("You must be logged in to create data", 401);
         }
@@ -537,7 +536,7 @@ class TalksController extends ApiController
 
         $data = $this->getLinkUserDataFromRequest($request);
 
-        if ($data['display_name'] == '' || $data['username'] == '') {
+        if ($data['display_name'] === '' || $data['username'] === '') {
             throw new Exception("You must provide a display name and a username", 400);
         }
 
@@ -546,11 +545,9 @@ class TalksController extends ApiController
 
         $claim = $talk_mapper->getSpeakerFromTalk($talk_id, $data['display_name']);
 
-        if (! $claim) {
+        if ($claim === false) {
             throw new Exception("No speaker matching that name found", 422);
-        }
-
-        if ($claim && $claim['speaker_id'] != null) {
+        } elseif ($claim['speaker_id'] != null) {
             throw new Exception("Talk already claimed", 422);
         }
 
@@ -575,7 +572,7 @@ class TalksController extends ApiController
         exit;
     }
 
-    protected function getLinkUserDataFromRequest(Request $request)
+    private function getLinkUserDataFromRequest(Request $request)
     {
         $talk = [];
         $talk['display_name'] = trim($request->getParameter('display_name', ''));
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -506,4 +506,60 @@ class TalksController extends ApiController
 
         return $talk;
     }
+
+    public function getSpeakersForTalk(Request $request, PDO $db)
+    {
+        $talk_id = $this->getItemId($request);
+        $talk = $this->getTalkById($db, $request, $talk_id);
+        $collection = new TalkModelCollection([$talk], 1);
+        return $collection->getTalks()[0]->speakers;
+
+
+    }
+
+    public function linkUserToTalk(Request $request, PDO $db)
+    {
+        //We must be logged in
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to create data", 400);
+        }
+
+        $talk_id = $this->getItemId($request);
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $user_id = $request->user_id;
+        $user_mapper = new UserMapper($db,$request);
+        $user = $user_mapper->getUserById($user_id)['users'][0];
+
+        $data = $this->getLinkUserDataFromRequest($request);
+
+        if ($data['display_name'] == '' || $data['username'] == ''){
+            throw new Exception("You must provide a display name and a username",400);
+        }
+
+        //Is the speaker this user?
+        if ($data['username'] === $user['username']){
+            echo "Speaker";
+        }elseif($talk_mapper->thisUserHasAdminOn($talk_id)){
+            echo "Event Admin";
+        }else{
+            throw new Exception("You must be the speaker or event admin to link a user to a talk", 400);
+        }
+
+
+    }
+
+    protected function getLinkUserDataFromRequest(Request $request)
+    {
+        $talk = [];
+        $talk['display_name'] = trim($request->getParameter('display_name',''));
+        $talk['username'] = trim($request->getParameter('username',''));
+
+        return $talk;
+
+    }
 }
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -498,11 +498,13 @@ class TalksController extends ApiController
             FILTER_SANITIZE_URL
         );
 
-        $talk['speakers'] = array_map(function ($speaker) {
-            $speaker = filter_var($speaker, FILTER_SANITIZE_STRING);
-            $speaker = trim($speaker);
-            return $speaker;
-        }, (array) $request->getParameter('speakers'));
+        $talk['speakers'] = array_map(
+            function ($speaker) {
+                $speaker = filter_var($speaker, FILTER_SANITIZE_STRING);
+                $speaker = trim($speaker);
+                return $speaker;
+            }, (array) $request->getParameter('speakers')
+        );
 
         return $talk;
     }
@@ -513,8 +515,6 @@ class TalksController extends ApiController
         $talk = $this->getTalkById($db, $request, $talk_id);
         $collection = new TalkModelCollection([$talk], 1);
         return $collection->getTalks()[0]->speakers;
-
-
     }
 
     public function linkUserToTalk(Request $request, PDO $db)
@@ -527,7 +527,7 @@ class TalksController extends ApiController
         $talk_id = $this->getItemId($request);
         $talk_mapper = new TalkMapper($db, $request);
         $talk = $talk_mapper->getTalkById($talk_id);
-        if (!$talk) {
+        if (! $talk) {
             throw new Exception("Talk not found", 404);
         }
 
@@ -537,35 +537,37 @@ class TalksController extends ApiController
 
         $data = $this->getLinkUserDataFromRequest($request);
 
-        if ($data['display_name'] == '' || $data['username'] == ''){
+        if ($data['display_name'] == '' || $data['username'] == '') {
             throw new Exception("You must provide a display name and a username",400);
         }
 
         //Get the speaker record based on the display name - check if this is already claimed,
-        // and otherwise ID becaomes the claim_id
+        //otherwise ID becomes the claim_id
 
-        $claim = $talk_mapper->getSpeakerFromTalk($talk_id,$data['display_name']);
+        $claim = $talk_mapper->getSpeakerFromTalk($talk_id, $data['display_name']);
 
-        if (! $claim){
+        if (! $claim) {
             throw new Exception("No speaker matching that name found", 400);
         }
 
-        if ($claim && $claim['speaker_id'] != null){
+        if ($claim && $claim['speaker_id'] != null) {
             throw new Exception("Talk already claimed", 400);
         }
 
         $pending_talk_claim_mapper = new PendingTalkClaimMapper($db, $request);
 
         //Is the speaker this user?
-        if ($data['username'] === $user['username']){
-            $pending_talk_claim_mapper->claimTalkAsSpeaker($talk_id,$user_id,$claim['ID']);
-        }elseif($talk_mapper->thisUserHasAdminOn($talk_id)){
+        if ($data['username'] === $user['username']) {
+            $pending_talk_claim_mapper->claimTalkAsSpeaker($talk_id, $user_id, $claim['ID']);
+            //We need to send an email to the host asking for confirmation
+        } elseif ($talk_mapper->thisUserHasAdminOn($talk_id)) {
             $speaker_id = $user_mapper->getUserIdFromUsername($data['username']);
-            if (! $speaker_id){
+            if (! $speaker_id) {
                 throw new Exception("Specified user not found");
             }
-            $pending_talk_claim_mapper->assignTalkAsHost($talk_id,$speaker_id,$claim['ID'],$user_id);
-        }else{
+            $pending_talk_claim_mapper->assignTalkAsHost($talk_id, $speaker_id, $claim['ID'], $user_id);
+            //We need to send an email to the speaker asking for confirmation
+        } else {
             throw new Exception("You must be the speaker or event admin to link a user to a talk", 400);
         }
 
@@ -576,10 +578,8 @@ class TalksController extends ApiController
     protected function getLinkUserDataFromRequest(Request $request)
     {
         $talk = [];
-        $talk['display_name'] = trim($request->getParameter('display_name',''));
-        $talk['username'] = trim($request->getParameter('username',''));
-
+        $talk['display_name'] = trim($request->getParameter('display_name', ''));
+        $talk['username'] = trim($request->getParameter('username', ''));
         return $talk;
-
     }
 }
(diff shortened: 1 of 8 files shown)


heiglandreas:
Would it be possible to move the refactoring bits out to a different PR?

src/controllers/TalksController.php
```diff
@@ -498,12 +498,89 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        );
+
+        return $talk;
+    }
+
+    public function getSpeakersForTalk(Request $request, PDO $db)
+    {
+        $talk_id = $this->getItemId($request);
+        $talk = $this->getTalkById($db, $request, $talk_id);
+        $collection = new TalkModelCollection([$talk], 1);
```
heiglandreas: What do you need the ```TalkModelCollection``` for? You pass it a TalkModel-instance to then retrieve that instance from it. Wouldn't a ```return $talk->speakers``` be sufficient?

src/controllers/TalksController.php
```diff
@@ -498,12 +498,89 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        return $collection->getTalks()[0]->speakers;
+    }
+
+    public function linkUserToTalk(Request $request, PDO $db)
```
heiglandreas: I'd name it ```setSpeakerForTalk``` in accordance with ```getSpeakersForTalk``` and to distinguish it from editing speakers for talks.

heiglandreas: Alternately claiming a talk is a two step-process with a user claiming a talk and an admin approving that claim. Why not move that to two separate methods like ```claimTalk``` and ```approveClaim```?

src/controllers/TalksController.php
```diff
@@ -498,12 +498,89 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        }
+
+        if ($claim && $claim['speaker_id'] != null) {
+            throw new Exception("Talk already claimed", 422);
+        }
 
+        $pending_talk_claim_mapper = new PendingTalkClaimMapper($db, $request);
+
+        //Is the speaker this user?
+        if ($data['username'] === $user['username']) {
```
heiglandreas: I didn't test this, but what happens when the current user is the event-admin and wants to claim his talk on the event? Happens sometimes in our usergroup. I know it's an edge case, but it wouldn't be possible to approve my own claim for a talk with this setup, wouldn't it?

src/controllers/TalksController.php
```diff
@@ -498,12 +498,89 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+            $pending_talk_claim_mapper->claimTalkAsSpeaker($talk_id, $user_id, $claim['ID']);
+            //We need to send an email to the host asking for confirmation
+        } elseif ($talk_mapper->thisUserHasAdminOn($talk_id)) {
+            $speaker_id = $user_mapper->getUserIdFromUsername($data['username']);
+            if (! $speaker_id) {
+                throw new Exception("Specified user not found", 404);
+            }
+            $pending_talk_claim_mapper->assignTalkAsHost($talk_id, $speaker_id, $claim['ID'], $user_id);
+            //We need to send an email to the speaker asking for confirmation
```
heiglandreas: Can't we trigger an event here that then takes care of notification or whatever else needs to be done? I know it's  not available ATM but wouldn't it be an idea to implement it at some point? And if so, why not for this feature?

src/models/PendingTalkClaimMapper.php
```diff
@@ -0,0 +1,102 @@
...
+     * @param int $claim_id   The ID from the talk_speaker table relating to the talk and display_name
+     *
+     * @return bool
+     */
+    public function claimTalkAsSpeaker($talk_id, $speaker_id, $claim_id)
+    {
+        $sql  = 'insert into pending_talk_claims 
+                    (talk_id,submitted_by,speaker_id,date_added,claim_id,user_approved_at) 
+                  values
+                    (:talk_id,:submitted_by,:speaker_id,UNIX_TIMESTAMP(),:claim_id,NOW())
```
heiglandreas: I'd rather not depend on the server-side settings for datetime- or timestamp-values but as we do that at different other places as well, let's be consistent :wink:

src/models/PendingTalkClaimMapper.php
```diff
@@ -0,0 +1,102 @@
...
+     * Propose a talk relationship by the host
+     *
+     * @param int $talk_id    The ID of the talk to claim
+     * @param int $speaker_id The ID of the speaker who owns the talk
+     * @param int $claim_id   The ID from the talk_speaker table relating to the talk and display_name
+     * @param int $user_id    The ID of the user proposing the relationship
+     *
+     * @return bool
+     */
+    public function assignTalkAsHost($talk_id, $speaker_id, $claim_id, $user_id)
```
heiglandreas: Wouldn't ```approveTalkClaim``` be more intuitive to understand what the method does? Especially as we can't assign a talk directly to a user without them claiming first.

src/controllers/TalksController.php
```diff
@@ -498,12 +498,89 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        //otherwise ID becomes the claim_id
+
+        $claim = $talk_mapper->getSpeakerFromTalk($talk_id, $data['display_name']);
+
+        if (! $claim) {
+            throw new Exception("No speaker matching that name found", 422);
+        }
+
+        if ($claim && $claim['speaker_id'] != null) {
+            throw new Exception("Talk already claimed", 422);
```
heiglandreas: This brings me to an interesting question: Can we "unclaim" a talk? Is that needed?

src/models/TalkMapper.php
```diff
@@ -772,4 +772,16 @@ public function delete($talk_id)
...
+
+    public function getSpeakerFromTalk($talk_id, $display_name)
+    {
+        $speaker_sql  = 'select ts.* from talk_speaker ts '
+            . 'where ts.talk_id = :talk_id and ts.speaker_name = :display_name';
+        $speaker_stmt = $this->_db->prepare($speaker_sql);
+        $speaker_stmt->execute(array("talk_id" => $talk_id, "display_name" => $display_name));
+        $speakers = $speaker_stmt->fetchAll(PDO::FETCH_ASSOC);
+
+        return count($speakers) > 0 ? $speakers[0] : false;
```
heiglandreas: Wouldn't it be easier to throw an exception when no speaker was found? That way you either receive an Exception from this method or an array containing user-values. So you wouldn't need to check the return-value for a boolean ```false``` every time. That might then refactor https://github.com/joindin/joindin-api/pull/352/files#diff-e27b217374c8436f4b830c74ed46cbbaR550

src/controllers/TalksController.php
liam-wiltshire: I did spend quite a bit of time over the naming of this - as per my message in IRC, either party can 'start' the linking process (the speaker can claim or the host can assign) - as such, I do think this 'setSpeakerForTalk' is a better name

src/controllers/TalksController.php
liam-wiltshire: No idea, I'm pleading tiredness :-)

src/controllers/TalksController.php
liam-wiltshire: https://joindin.jira.com/browse/JOINDIN-643 :-)

src/controllers/TalksController.php
liam-wiltshire: As per IRC, this endpoint is only the first 'half' of the talk->speaker link process - as such, if you are the host and the speaker, this would claim the talk as you, the speaker, which would then show up as pending approval for you, the host, which would use a separate (and as yet unwritten) endpoint.

src/models/PendingTalkClaimMapper.php
liam-wiltshire: As per IRC, that's not really what this does - approving will be a seperate endpoint that isn't written yet. This is for the host taking the first step, and saying which user was the speaker, so I think `assignTalkAsHost` is about as descriptive as we can be,

src/models/PendingTalkClaimMapper.php
liam-wiltshire: Makes no odds to me either way - we use `NOW()` fairly extensively,  and realistically if the server time is wrong, it's going to make no difference if we do it in PHP, SQL or Bash anyway :-)

src/models/TalkMapper.php
liam-wiltshire: I see where you are coming from, but I'm not sure that's the right approach in a mapper - I would always expect a getter in a mapper to return - either the data exists, or it doesn't.

Certainly we don't throw exceptions anywhere else in getters in the mappers - we do on set/create methods, and I think that makes sense, but this would be the 'odd one out' otherwise?

src/controllers/TalksController.php
liam-wiltshire: We probably should do it, however I'm not sure it should be part of this scope - I certainly think someone else (you? :-)) could do a better job of it, as event handling isn't something I've done much of - I've used symfony/event-dispatcher a couple of times and that's about it!

If someone was to submit a PR with the basis of an event system however, I'd happily then start updating endpoints to use them!

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+            throw new Exception("Talk not found", 404);
+        }
+
+        $user_id = $request->user_id;
+        $user_mapper = new UserMapper($db, $request);
+        $user = $user_mapper->getUserById($user_id)['users'][0];
+
+        $data = $this->getLinkUserDataFromRequest($request);
+
+        if ($data['display_name'] == '' || $data['username'] == '') {
```
asgrim: Ideally strict type checks (`===`) should be used. If you're unsure of the data type of the values, cast them to a `string` (it's equivalent in logic, but you're being explicit about expectations which improves readability)

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        //We must be logged in
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to create data", 401);
+        }
+
+        $talk_id = $this->getItemId($request);
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (! $talk) {
+            throw new Exception("Talk not found", 404);
```
asgrim: Not sure what the policy is for Joind.in regarding this, but I'd encourage using specific exception classes - at minimum, the SPL exceptions could be used instead. Apologies if this is an already-discussed point, or if there is an import that I missed

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+    public function getSpeakersForTalk(Request $request, PDO $db)
+    {
+        $talk_id = $this->getItemId($request);
+        $talk = $this->getTalkById($db, $request, $talk_id);
+        return $talk->speakers;
+    }
+
+    public function setSpeakerForTalk(Request $request, PDO $db)
+    {
+        //We must be logged in
```
asgrim: IMO this comment is unnecessary - it's clear what is going on when reading the code

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+        if ($data['display_name'] == '' || $data['username'] == '') {
+            throw new Exception("You must provide a display name and a username", 400);
+        }
+
+        //Get the speaker record based on the display name - check if this is already claimed,
+        //otherwise ID becomes the claim_id
+
+        $claim = $talk_mapper->getSpeakerFromTalk($talk_id, $data['display_name']);
+
+        if (! $claim) {
```
asgrim: What type is `$claim`? If it's a `bool` this is fine, but looking at the assignment, it comes from `getSpeakerFromTalk` which suggests some kind of data. Therefore this comparison should be tightened up (e.g. does `getSpeakerFromTalk` return `null` or `false` etc. on failure?)

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+            throw new Exception("No speaker matching that name found", 422);
+        }
+
+        if ($claim && $claim['speaker_id'] != null) {
```
asgrim: The first check (`if ($claim)`) is not needed any more as you checked it a few lines up with the exception (so execution would not reach here anyway)

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+                throw new Exception("Specified user not found", 404);
+            }
+            $pending_talk_claim_mapper->assignTalkAsHost($talk_id, $speaker_id, $claim['ID'], $user_id);
+            //We need to send an email to the speaker asking for confirmation
+        } else {
+            throw new Exception("You must be the speaker or event admin to link a user to a talk", 401);
+        }
+
+        header("Location: " . $request->base . $request->path_info, null, 204);
+        exit;
```
asgrim: Is exit intended here? Seems rather abrupt, and untestable?

src/controllers/TalksController.php
```diff
@@ -498,12 +498,88 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+    }
+
+    protected function getLinkUserDataFromRequest(Request $request)
```
asgrim: Does this need to be `protected`? Seems like it should be `private` to me maybe?

src/models/PendingTalkClaimMapper.php
```diff
@@ -0,0 +1,102 @@
...
+{
+
+    /**
+     * Default mapping for column names to API field names
+     *
+     * @return array with keys as API fields and values as db columns
+     */
+    public function getDefaultFields()
+    {
+        $fields = array(
```
asgrim: No short array syntax? Looks like `composer.lock` requires min PHP of `>=5.4.0` (though, oddly, it's not mentioned in `composer.json`), so short syntax is available.

src/models/TalkModel.php
```diff
@@ -85,6 +85,7 @@ public function getOutputView(Request $request, $verbose = false)
         $item['verbose_comments_uri'] = $base . '/' . $version . '/talks/' . $this->ID
                                         . '/comments?verbose=yes';
         $item['event_uri']            = $base . '/' . $version . '/events/' . $this->event_id;
+        $item['speakers_uri']          = $base . '/' . $version . '/talks/' . $this->ID . '/speakers';
```
asgrim: Alignment looks off here

src/controllers/TalksController.php
liam-wiltshire: I'm in two minds about this - I totally get where you are coming from, but in 99% of cases the API throws `\Exception` - I don't think it's a policy per-sey, more that they are not really 'meant' to be caught - the handle_exception catches all of them and then turns it into an API response.

That said, we have been discussing the need to standardise the HTTP codes we return, so perhaps that could be tied in with creating our own Exception classes, where the HTTP code is pre-defined as part of the exception. 

This would probably be a separate scope to this PR, but I'll look at raising a separate PR to do it.

src/controllers/TalksController.php
liam-wiltshire: They are strings anyway, so it doesn't need casting.

src/controllers/TalksController.php
liam-wiltshire: Yes, again there are probably better ways to do it, but this is the way it's done throughout the codebase. You are right it causes testing issues (hence I did this - https://github.com/joindin/joindin-api/pull/342/files#diff-000feae194979d9dea78c415431bef50R324 - which I'm certainly not proud of!), and it probably needs looking at within the larger scope of tidying up the APIs at some point

src/controllers/TalksController.php
heiglandreas: I've just created https://joindin.jira.com/browse/JOINDIN-708 to at least not forget that point!

src/models/PendingTalkClaimMapper.php
heiglandreas: I was more referring to the timezone settings of the server. But as long as the times are consistent within the database, it doesn't matter.

src/models/PendingTalkClaimMapper.php
heiglandreas: Thanks for the clarification on IRC! I now absolutely understand that!

src/models/TalkMapper.php
heiglandreas: Let's stay consistent!

src/controllers/TalksController.php
```diff
@@ -498,12 +498,85 @@ protected function getTalkDataFromRequest(PDO $db, Request $request, $event_id)
...
+    }
+
+    public function getSpeakersForTalk(Request $request, PDO $db)
+    {
+        $talk_id = $this->getItemId($request);
+        $talk = $this->getTalkById($db, $request, $talk_id);
+        return $talk->speakers;
+    }
+
+    public function setSpeakerForTalk(Request $request, PDO $db)
```
carnage: This seems like a lot of work to do in a controller method. You have access checks, validation, filtering, creation of service classes and DB interaction all mixed into one. This will be a nightmare for testing and future maintenance.

Would it be possible to separate some of this logic out into separate methods or ideally separate classes?

src/controllers/TalksController.php
liam-wiltshire: https://joindin.jira.com/browse/JOINDIN-709 :-)

src/controllers/TalksController.php
liam-wiltshire: I understand what you are saying, the but the API design does lend itself to 'fat controllers' rather - see https://github.com/joindin/joindin-api/blob/master/src/controllers/TalksController.php#L48 as an example, or https://github.com/joindin/joindin-api/blob/master/src/controllers/EventsController.php#L149

liam-wiltshire:
Unit Tests added

src/controllers/TalksController.php
carnage: I don't think there is any reason why we shouldn't seek to improve that design. I'll PR a couple of suggestions.

PR #356, opened 2016-10-08
Approve talk claim

liam-wiltshire:


liam-wiltshire:
Something funky happened with that rebase....

PR #358, opened 2016-10-08
Adds endpoint and functionality to remove a speaker from a talk

railto:
Fixes #JOINDIN-643

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -654,7 +654,7 @@ class TalksController extends ApiController
         }
 
         // delete speaker from talk
-        $talk_mapper->removeSpeakerFromTalk($talk_id, $speaker_id);
+        $talk_mapper->removeApprovedSpeakerFromTalk($talk_id, $speaker_id);
 
         $uri = $request->base . '/' . $request->version . '/talks/' . $talk_id;
         header('Location: ' . $uri, null, 204);
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -630,7 +630,7 @@ class TalksController extends ApiController
     public function removeSpeakerFromTalk(Request $request, PDO $db)
     {
         if (!isset($request->user_id)) {
-            throw new Exception("You must be logged in to delete data", 400);
+            throw new Exception("You must be logged in to delete data", 401);
         }
 
         $talk_id = $this->getItemId($request);
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -644,13 +644,13 @@ class TalksController extends ApiController
 
         $speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $speaker_id);
         if (!$speaker) {
-            throw new Exception("Provided user is not a speaker on this talk", 400);
+            throw new Exception("Provided user is not a speaker on this talk", 404);
         }
 
         $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
         $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
         if (!($is_admin || $is_speaker)) {
-            throw new Exception("You do not have permission to remove this speaker from this talk", 400);
+            throw new Exception("You do not have permission to remove this speaker from this talk", 403);
         }
 
         // delete speaker from talk
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -627,7 +627,7 @@ class TalksController extends ApiController
         return $this->pending_talk_claim_mapper;
     }
 
-    public function removeSpeakerFromTalk(Request $request, PDO $db)
+    public function removeApprovedSpeakerFromTalk(Request $request, PDO $db)
     {
         if (!isset($request->user_id)) {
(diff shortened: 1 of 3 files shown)


liam-wiltshire:
ok to test

src/controllers/TalksController.php
```diff
@@ -626,4 +626,38 @@ public function getPendingTalkClaimMapper($db, $request)
 
         return $this->pending_talk_claim_mapper;
     }
+
+    public function removeSpeakerFromTalk(Request $request, PDO $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 400);
```
heiglandreas: Please use a 401 (Unauthorized) here :wink:

src/controllers/TalksController.php
```diff
@@ -626,4 +626,38 @@ public function getPendingTalkClaimMapper($db, $request)
...
+
+        $speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $speaker_id);
+        if (!$speaker) {
+            throw new Exception("Provided user is not a speaker on this talk", 400);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        $is_speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $request->user_id);
+        if (!($is_admin || $is_speaker)) {
+            throw new Exception("You do not have permission to remove this speaker from this talk", 400);
```
heiglandreas: And this is a case for a 403 (Forbidden)

src/controllers/TalksController.php
```diff
@@ -626,4 +626,38 @@ public function getPendingTalkClaimMapper($db, $request)
...
+
+        $talk_mapper = new TalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $speaker = $talk_mapper->isUserASpeakerOnTalk($talk_id, $speaker_id);
+        if (!$speaker) {
+            throw new Exception("Provided user is not a speaker on this talk", 400);
```
heiglandreas: Wouldn't this qualify for a 404 (Not found)?

src/models/TalkMapper.php
```diff
@@ -784,4 +784,16 @@ public function getSpeakerFromTalk($talk_id, $display_name)
...
     }
+
+    public function removeSpeakerFromTalk($talk_id, $speaker_id)
+    {
+        $params = [
+            'talk_id' => $talk_id,
+            'speaker_id' => $speaker_id,
+        ];
+
+        $sql = 'update talk_speaker set speaker_id = null where talk_id = :talk_id and speaker_id = :speaker_id';
```
heiglandreas: Am I right in thinking that this removal only works when a speaker has claimed the talk and the claim was approved? So removing a speaker that has not yet claimed the talk doesn't work currently?

liam-wiltshire: @heiglandreas - yes, that is already done - if you are a host, and you want to completely remove a speaker from a talk (not just unlink it), then that is done as part of editing a talk, which is already live.

src/controllers/TalksController.php
railto: Good point, changing.

src/models/TalkMapper.php
railto: Good point, changing.

heiglandreas:
Looks good from my POV.

heiglandreas:
Sorry. but on testing I found 2 things:

1. calling ```http://api.dev.joind.in/v2.1/talks/[\d+]/``` gives the speakers_uri ```http://api.dev.joind.in/v2.1/talks/[\d+]/speakers``` (note the *s* at the end). But the route in ```routes/2.1.config``` talks about ```speaker``` (note the missing *s*). Which causes an "Operation not permitted, Sorry"-Exception... So please add an *s* there ;)

2. This endpoint only removes the claim, not the speaker from the talk.The call to  ```DELETE http://api.dev.joind.in/v2.1/talks/[\d+]/speaker[s]/[\d+]``` returns a location-header that points to ```http://api.dev.joind.in/v2.1/talks/[\d+]``` which then returns the list of speakers that still contains the name of the just deleted speaker. So after removing the claim the speaker has to be removed separately from the talk using the talk-edit-Endpoint. I'd expect the user completely to be removed from a talk after calling **DELETE** on a speaker-entry. Am I the only one?

railto:
@heiglandreas Point 1 corrected. As for point 2, the original spec for this in JIRA called for only the speaker_id column to be changed with the remainder of the row to remain untouched.  I personally think that it should also remove the name but coded to the spec.

liam-wiltshire:
@railto @heiglandreas - as per IRC I accidentally merged it when I was trying to comment - stupid mobile layout...

However I've had a look through the code, and functionally I think it's all fine - does need Unit Tests, but as discussed these can follow.

My only concern is that the method name still don't 'feel' right to me, however I also can't think of a better name.

As such, I think it's not worth reverting the PR - once my PR for approving claims is merged it's going to need some in-depth testing anyway as we've made some fairly sizable changes, so my view is that we go through a proper test once all the functionality is in before we do a deploy to production again.

Thanks,

Liam

PR #363, opened 2016-10-12
Fixing issue where speaker_id = 0

liam-wiltshire:
There was an issue where (at least in the test data) a speaker_id of 0 was sometimes used instead of NULL in the talk_speaker table. Because there was only a check for null, the API was reporting that these talks were claimed when that wasn't the case.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -548,7 +548,7 @@ class TalksController extends ApiController
 
         if ($claim === false) {
             throw new Exception("No speaker matching that name found", 422);
-        } elseif ($claim['speaker_id'] != null) {
+        } elseif ($claim['speaker_id'] != null && $claim['speaker_id'] != 0) {
             throw new Exception("Talk already claimed", 422);
         }
 


src/controllers/TalksController.php
```diff
@@ -548,7 +548,7 @@ public function setSpeakerForTalk(Request $request, PDO $db)
 
         if ($claim === false) {
             throw new Exception("No speaker matching that name found", 422);
-        } elseif ($claim['speaker_id'] != null) {
+        } elseif ($claim['speaker_id'] != null && $claim['speaker_id'] != 0) {
```
heiglandreas: Shouldn't that be the same due to not using ```!==``` and the internal type-conversion?

liam-wiltshire: It should, but for some reason (that I've not been able to figure out) it's not.... with it was just != null, rows with a speaker_id of 0 were throwing the 'already claimed' exception, and putting in the != 0 worked...

I'm at a loss to explain why, but that's what it's doing!

PR #364, opened 2016-10-12
Talk claim emails

liam-wiltshire:
Generation and sending of required emails for new world talk claims.

Updated unit tests due to additional dependency (event mapper)

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -593,7 +593,10 @@ class TalksController extends ApiController
                 } else {
                     throw new Exception("There was a problem assigning the talk", 500);
                 }
-                //We need to send an email to the speaker asking for confirmation
+
+                $recipients   = [$user_mapper->getEmailByUserId($speaker_id)];
+                $emailService = new TalkClaimApprovedEmailService($this->config, $recipients, $event, $talk);
+                $emailService->sendEmail();
             } else {
                 throw new Exception("You must be an event admin to approve this claim", 401);
             }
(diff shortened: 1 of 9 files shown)


src/controllers/TalksController.php
```diff
@@ -632,6 +653,21 @@ public function getTalkMapper($db, $request)
         return $this->talk_mapper;
     }
 
+    public function setEventMapper(EventMapper $event_mapper)
```
heiglandreas: Just an idea as I'm seeing a lot of setters and getters sprouting over the next commits: Can't we use Factory-Classes for getting the mappers? Either the default mapper is instantiated or you can call a static setter and then the set mapper will be returned. Basically moving the setMapper/getMapper into its own class.

src/models/EventMapper.php
```diff
@@ -453,6 +453,8 @@ public function transformResults($results, $verbose)
                             . $row['ID'] . '/comments/reported';
                         $list[$key]['reported_talk_comments_uri'] = $base . '/' . $version . '/events/'
                             . $row['ID'] . '/talk_comments/reported';
+                        $list[$key]['pending_claims_uri'] = $base . '/' . $version . '/events/'
```
heiglandreas: Is it possible to show this URI only to people that actually can do something with it? So AFAIK that'd be event-admins. I find it nicer to only show people stuff they actually can use. And most people will get a 403 (or 401) when calling that URI.

railto: I agree that this would be preferable, let's not show stuff that the majority of people can't access

liam-wiltshire: @railto - as per comment on  other PR, I'm pretty sure it's inside the can Edit conditional? (Not at pc to check )

liam-wiltshire: Yes it is - see https://github.com/liam-wiltshire/joindin-api/blob/e9839cf7164e0388d8cb413959ddf9803ff399f4/src/models/EventMapper.php#L451

Issue #357, opened 2016-10-08
Creating endpoint to allow users to change their username


liam-wiltshire:
Something that users' can't currently do is change their username. It's been asked for on Twitter, and while it's a feature that probably wouldn't be used often, when users do want to change their username it's normally for a very good reason.

I can't see any reason why username needs to be immutable (but I'm sure @akrabat  would be better placed to confirm), so we need either need a modification of the PUT user endpoint to allow the username field, or a new endpoint to change a username separately from editing users. 

Ideally, this should also be recorded in a lookup table of some description, so that if someone has a link containing the old username we can link it to the new username, but this is less critical for now I think.

---

Joind.In uses JIRA for managing issues - please claim at: https://joindin.jira.com/browse/JOINDIN-696

runz0rd:
Im on it. Where exactly would the lookup table be? Do you have anything like it implemented?

liam-wiltshire:
This is a pretty new feature, we've not really done anything similar.

It could probably be just as straightforward as a new table which contains the user_id and the old username - we will need to update other API calls to then look up old usernames, but that can be done separately.

We use DB patches for stuff like that - see https://github.com/joindin/joindin-api#database-patches

runz0rd:
Ill take a look, thanks!

dstockto:
One thing to consider if allowing a user to essentially hold onto the old username (via the redirect) is whether there's a limit on the number of renames. For one the redirect chain could get long, and secondly, it introduces a potential for a DOS of sorts on usernames. Of course, that's still technically possible now, but it would require someone to go and register a whole load of different users. Being able to rename over and over essentially removes a potential username from the pool of possible names and denies anyone else the ability to use it. I'm not saying to not allow rename or that this will even be a problem, I just wanted to mention that it could be a problem.

runz0rd:
This is definitely useful info. Do you guys want a limit for username renames?
 Also, "Being able to rename over and over essentially removes a potential username from the pool of possible names and denies anyone else the ability to use it." - I was not aware the old usernames of one user would prevent other users from taking that username. Do you want this functionality?

liam-wiltshire:
@runz0rd well, yes - if we are having a lookup table that allows requests to the old username to be re-routed to the new username, then by extension, that username can't be re-used for someone else.

[...]

PR #365, opened 2016-10-13
implemented username updating inside the updateUser method

runz0rd:
For issue #357
JIRA JOINDIN-696

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -247,6 +247,22 @@ class UsersController extends ApiController
                 }
             }
 
+            $username = $request->getParameter("username", false);
+            if (false !== $username) {
+                // does anyone else have this username?
+                $existing_user = $user_mapper->getUserByUsername($username);
+                if ($existing_user['users']) {
+                    // yes but is that our existing user being found?
+                    $old_user = $user_mapper->getUserById($userId);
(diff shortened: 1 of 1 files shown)


liam-wiltshire:
ok to test

liam-wiltshire:
Thanks for this - I'll try to pull it down and merge it this evening, if not tomorrow

runz0rd:
No problem :+1:

src/controllers/UsersController.php
```diff
@@ -247,6 +247,22 @@ public function updateUser(Request $request, $db)
...
+                if ($existing_user['users']) {
+                    // yes but is that our existing user being found?
+                    $old_user = $user_mapper->getUserById($userId);
+                    if ($old_user['users'][0]['uri'] != $existing_user['users'][0]['uri']) {
+                        // the username exists and not on this user's account
+                        $errors[] = "That username is already associated with another account";
+                    }
+                }
+
+                $user['username'] = filter_var(trim($username), FILTER_SANITIZE_STRING);
```
heiglandreas: Shouldn't that happen right after the ```if (false !== $username) {```?

This is checking an unsanitized username against the database-entries but that username might be changed when sanitized. So IMO the checks should be done **after** sanitizing the username

runz0rd:
You're right. I'll make another pull request with the changes.

PR #367, opened 2016-10-15
Pending talk claims

liam-wiltshire:
Create endpoint for pending talk claims

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -529,6 +529,24 @@ class EventsController extends ApiController
         }
     }
 
+    public function pendingClaims($request, $db)
+    {
+        // Check for login
+        if (! isset($request->user_id)) {
+            throw new Exception("You must be logged in to view pending claims", 401);
+        }
+
+        $event_id = $this->getItemId($request);
+        $event_mapper = $this->getEventMapper($db, $request);
+
+        $pending_talk_claim_mapper = $this->getPendingTalkClaimMapper($db, $request);
+        if (! $event_mapper->thisUserHasAdminOn($event_id)) {
+            throw new Exception('You do not have permission to edit this track', 403);
+        }
+
+        return $pending_talk_claim_mapper->getPendingClaimsByEventId($event_id);
+    }
+
     /**
      * Create track
      *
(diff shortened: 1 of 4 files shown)


src/controllers/EventsController.php
```diff
@@ -666,4 +684,33 @@ public function rejectAction($request, $db)
...
+    }
+
+    public function getPendingTalkClaimMapper($db, $request)
+    {
+        if (! isset($this->pending_talk_claim_mapper)) {
+            $this->pending_talk_claim_mapper = new PendingTalkClaimMapper($db, $request);
+        }
+
+        return $this->pending_talk_claim_mapper;
+    }
```
heiglandreas: just an idea as I'm seeing a lot of setters and getters sprouting over the next commits: Can't we use Factory-Classes for getting the mappers? Either the default mapper is instantiated or you can call a static setter and then the set mapper will be returned. Basically moving the setMapper/getMapper into its own class.

heiglandreas: And where is the test using it? 😛

src/models/PendingTalkClaimMapper.php
```diff
@@ -167,4 +167,40 @@ public function approveClaimAsHost($talk_id, $speaker_id, $claim_id)
...
+                    'talk_uri'          => $base . '/' . $version . '/talks/' . $row['talk_id'],
+                    'speaker_uri'       => $base . '/' . $version . '/users/' . $row['speaker_id'],
+                    'approve_claim_uri' => ''
+                ];
+
+            }
+
+            return $retval;
+
+        } else {
```
heiglandreas: This else can be omitted.

src/models/PendingTalkClaimMapper.php
```diff
@@ -167,4 +167,40 @@ public function approveClaimAsHost($talk_id, $speaker_id, $claim_id)
...
+
+        $sql = 'select c.*, s.speaker_name from pending_talk_claims c
+                    inner join talks t on t.ID = c.talk_id
+                    inner join talk_speaker s on s.ID = c.claim_id
+                    where t.event_id = :event_id and
+                    (host_approved_at IS NULL or user_approved_at IS NULL)';
+        $stmt = $this->_db->prepare($sql);
+        $response = $stmt->execute(['event_id' => $event_id]);
+
+        if ($response && $stmt->rowCount() > 0) {
```
heiglandreas: I'd alter this to
 ```php
if (! $response || $stmt->rowCount() < 1) {
    return false;
}
```
and then continue with the stuff you currently have in the else.

That way the "default flow" is going through linearily and exceptions fail fast

File: src/models/EventMapper.php [hunk 4]
```diff
@@ -453,6 +453,8 @@ public function transformResults($results, $verbose)
                             . $row['ID'] . '/comments/reported';
                         $list[$key]['reported_talk_comments_uri'] = $base . '/' . $version . '/events/'
                             . $row['ID'] . '/talk_comments/reported';
+                        $list[$key]['pending_claims_uri'] = $base . '/' . $version . '/events/'
```
liam-wiltshire: Erm... it is - check line 451 `if ($this->thisUserHasAdminOn($row['ID'])) {`

:-)

src/controllers/EventsController.php
liam-wiltshire: I agree, however I don't think it should form part of this PR. With ZendCon this week, I think we are well and truly in 'Conference Season', so I think getting this claim function live needs to be a priority.

That said, if we are going to aim to get test coverage up, we certainly need to do something, either DI or Factories - fancy raising a JIRA ticket? :-)

src/models/EventMapper.php
heiglandreas: Ouch! That happens when you only review the changed code without having a look at the surrounding stuff....

Sorry 'bout that and forget my comment!

Issue #361, opened 2016-10-10
Endpoint to allow a comment to be edited


liam-wiltshire:
At the moment, if you post a comment, and then notice a typo etc there is no way of editing it.

We would like an endpoint that allows only the author of the comment, (not a host or a site admin) to send an update to the comment.

This should be a PUT against a comment (e.g. http://test.api.joind.in/v2.1/talk_comments/72870)

As a point of validation, comments should only be editable for 15 minutes after creation - we might change this in future, so it would be good if this value was stored in configuration.

---

Joind.In uses JIRA for managing issues - please claim at: https://joindin.jira.com/browse/JOINDIN-686

andela-celisha-wigwe:
Hey @liam-wiltshire, I would like to try this out.

liam-wiltshire:
Awesome! If you could please claim it on Jira, that'd be great.

runz0rd:
@andela-celisha-wigwe Are you making progress on this? If not, mind if I take this?

andela-celisha-wigwe:
@runz0rd Please feel free to work on it.

runz0rd:
@liam-wiltshire 
Is the controller method for this feature supposed to redirect after success, and where to? If not what should it return?
As for the time limit I was thinking of storing it in config['limits'] and using a hardcoded default of 15 mins if the configuration value has not been set.
Also I noticed you dont do any filtering on the comment body in your other methods, so I left it as is.

liam-wiltshire:
The API endpoint doesn't handle redirects, so it should return an error code on fail (throwing a redirect handles that). Otherwise, on success a no content  response is fine.


Sent from my Samsung Galaxy smartphone.
-------- Original message --------From: Milos Pejanovic <notifications@github.com> Date: 16/10/2016  18:31  (GMT+00:00) To: joindin/joindin-api <joindin-api@noreply.github.com> Cc: Liam Wiltshire <liam@w.iltshi.re>, Mention <mention@noreply.github.com> Subject: Re: [joindin/joindin-api] Endpoint to allow a comment to be edited (#361) 
@liam-wiltshire 

Is the controller method for this feature supposed to redirect after success, and where to? If not what should it return?

As for the time limit I was thinking of storing it in config['limits'] and using a hardcoded default of 15 mins if the configuration value has not been set.

Also I noticed you dont do any filtering on the comment body in your other methods, so I left it as is.

—
You are receiving this because you were mentioned.
[...]

PR #368, opened 2016-10-16
Edit talk comment

runz0rd:
For #361 and https://joindin.jira.com/browse/JOINDIN-686

### DIFF
--- a/src/controllers/Talk_commentsController.php
+++ b/src/controllers/Talk_commentsController.php
@@ -133,8 +133,6 @@ class Talk_commentsController extends ApiController
 
     public function updateComment($request, $db)
     {
-        $request->user_id = 14;
-
         // must be logged in
         if (!isset($request->user_id) || empty($request->user_id)) {
             throw new Exception('You must log in to edit a comment', 401);
--- a/src/controllers/Talk_commentsController.php
+++ b/src/controllers/Talk_commentsController.php
@@ -130,4 +130,39 @@ class Talk_commentsController extends ApiController
         header("Location: $uri", true, 204);
         exit;
     }
+
+    public function updateComment($request, $db)
+    {
+        if (false == ($request->getUserId())) {
+            throw new Exception("You must be logged in to edit a comment", 401);
+        }
+
+        //TODO check if the logged user is the user editing the comment
+
+        $comment_id = $this->getItemId($request);
+        $new_comment_body = $request->getParameter('comment');
+        if (empty($new_comment_body)) {
+            throw new Exception('The field "comment" is required', 400);
+        }
+
+        $comment_mapper = new TalkCommentMapper($db, $request);
+        $comment_timestamp = $comment_mapper->getCommentTimestamp($comment_id);
+        if (false === $comment_timestamp) {
+            throw new Exception('Comment not found', 404);
+        }
+
+        $max_comment_edit_minutes = 15;
+        if (isset($this->config['limits']['max_comment_edit_minutes'])) {
+            $max_comment_edit_minutes = $this->config['limits']['max_comment_edit_minutes'];
+        }
+
+        if ($comment_timestamp + $max_comment_edit_minutes * 60 > time()) {
+            throw new Exception('Cannot edit the comment after ' . $max_comment_edit_minutes . ' minutes', 400);
+        }
+        $comment_mapper->updateCommentBody($comment_id, $new_comment_body);
+
+        $uri = $request->base  . '/' . $request->version . '/talk_comments/' . $comment_id;
+        header("Location: $uri", true, 204);
+        exit;
+    }
 }
--- a/src/controllers/Talk_commentsController.php
+++ b/src/controllers/Talk_commentsController.php
@@ -133,22 +133,28 @@ class Talk_commentsController extends ApiController
 
     public function updateComment($request, $db)
     {
-        if (false == ($request->getUserId())) {
-            throw new Exception("You must be logged in to edit a comment", 401);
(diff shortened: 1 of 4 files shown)


railto:
Ok to test

liam-wiltshire:
ok to test

runz0rd:
Want me to get those curly braces in the new line?

liam-wiltshire:
Please :-)

src/controllers/Talk_commentsController.php
```diff
@@ -130,4 +130,44 @@ public function moderateReportedComment($request, $db)
...
+        $comment_id = $this->getItemId($request);
+        $comment_mapper = new TalkCommentMapper($db, $request);
+        $comment = $comment_mapper->getRawComment($comment_id);
+
+        if (false === $comment) {
+            throw new Exception('Comment not found', 404);
+        }
+
+        if ($comment['user_id'] != $request->user_id) {
+            throw new Exception('You are not the comment author', 401);
```
heiglandreas: This should be a 403 as return code (HTTP-Status Forbidden) as we know the user but that user just isn't allowed doing this action.

src/controllers/Talk_commentsController.php
```diff
@@ -130,4 +130,44 @@ public function moderateReportedComment($request, $db)
...
+        $max_comment_edit_minutes = 15;
+        if (isset($this->config['limits']['max_comment_edit_minutes'])) {
+            $max_comment_edit_minutes = $this->config['limits']['max_comment_edit_minutes'];
+        }
+
+        if ($comment['date_made'] + ($max_comment_edit_minutes * 60) < time()) {
+            throw new Exception('Cannot edit the comment after ' . $max_comment_edit_minutes . ' minutes', 400);
+        }
+
+        $comment_mapper->updateCommentBody($comment_id, $new_comment_body);
```
heiglandreas: Are we sure that this will always work? I'd either check the return value or check for any thrown Exceptions within ```updateCommentBody```

src/models/TalkCommentMapper.php
```diff
@@ -386,4 +386,37 @@ public function moderateReportedComment($decision, $comment_id, $user_id)
...
+
+    /**
+     * @param integer $comment_id
+     * @param string $new_comment_body
+     */
+    public function updateCommentBody($comment_id, $new_comment_body)
+    {
+        $sql = "update talk_comments set comment = :new_comment_body where ID = :comment_id";
+        $stmt = $this->_db->prepare($sql);
+        $stmt->execute([
```
heiglandreas: Check for the return-value of the execute-command here please. Or return it. THat way the caller of the method gets an information whether the method succeeeded or failed and can react appropriately.

src/controllers/Talk_commentsController.php
```diff
@@ -130,4 +130,44 @@ public function moderateReportedComment($request, $db)
...
+            throw new Exception('You are not the comment author', 401);
+        }
+
+        $max_comment_edit_minutes = 15;
+        if (isset($this->config['limits']['max_comment_edit_minutes'])) {
+            $max_comment_edit_minutes = $this->config['limits']['max_comment_edit_minutes'];
+        }
+
+        if ($comment['date_made'] + ($max_comment_edit_minutes * 60) < time()) {
+            throw new Exception('Cannot edit the comment after ' . $max_comment_edit_minutes . ' minutes', 400);
```
liam-wiltshire: I'm torn here - technically a 403 would satisfy (in the same way as a 403 for trying to access a directory listing that is 'forbidden for all'), but it doesn't _feel_ right to me? I think 400 is ok, or 405, because PUT isn't allowed after the 15 minute window?

runz0rd: It makes sense to me that this remains a 400, but whatever you guys choose, Ill put in.

liam-wiltshire: Leave it as a 400 for now - we are looking to replace all the Exceptions anyway with descriptive exceptions, so it'll probably change some point fairly soon anyway.

liam-wiltshire:
Thanks @runz0rd

src/models/TalkCommentMapper.php
```diff
@@ -386,4 +386,40 @@ public function moderateReportedComment($decision, $comment_id, $user_id)
...
+    /**
+     * @param integer $comment_id
+     * @return false|array
+     */
+    public function getRawComment($comment_id)
+    {
+        $sql = "select tc.* from talk_comments tc where tc.ID = :comment_id";
+        $stmt = $this->_db->prepare($sql);
+        $response = $stmt->execute(array(':comment_id' => $comment_id));
+        if ($response) {
```
heiglandreas: I'd recode this as follows:

```php
if (! $response) {
    return false;
}

$result = $stmt...

if (! isset($results[0]) {
    return false;
}

return $results[0];
```

That way it is more clear what the expected path of action is and where the "failures" are.

But that's merely a personal thing.

src/controllers/Talk_commentsController.php
```diff
@@ -130,4 +130,47 @@ public function moderateReportedComment($request, $db)
         header("Location: $uri", true, 204);
         exit;
     }
+
+    public function updateComment($request, $db)
+    {
+        $request->user_id = 14;
```
liam-wiltshire: @runz0rd - I guess this was put in for testing? Needs removing before I can merge please :-)

runz0rd: Ah, sorry about that.

PR #369, opened 2016-10-27
Fix escaped single and double quotes in Talk-Informations

heiglandreas:
This is a test to verify that the API is not doing any stranfge escaping stuff and returns items as it receives them.

References https://github.com/joindin/joindin-web2/issues/378 and https://github.com/joindin/joindin-web2/issues/366


heiglandreas:
Please can someone re-check the DB-Patch to be absolutely sure that the DB will not be nuked? 😄

PR #370, opened 2016-10-27
Adds a way to retrieve the hosts of an event

heiglandreas:
This is the first PR to address https://joindin.jira.com/browse/JOINDIN-529

This one contains the `GET /events/[event-id]/hosts`-endpoint.

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -1,4 +1,9 @@
 [
+    {    "path": "/events(/[^/]+)*/hosts",
+        "controller": "Event_hostsController",
+        "action": "listHosts",
+        "verbs": ["GET"]
+    },
     {
         "path": "/events(/[^/]+)*/comments/reported",
         "controller": "Event_commentsController",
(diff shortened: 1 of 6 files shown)


liam-wiltshire:
@heiglandreas thanks for this.

However, shouldn't the hosts uri be exposed on the other get events endpoints?

For example, http://api.dev.joind.in/v2.1/events/45?verbose=yes has comments_uri, talks_uri etc, so it'd make sense for hosts_uri to be there too? We use these properties on web2 to then feed back into the API models.

Thanks!

heiglandreas:
You'Re welcome :wink:

Anywhere else than in the events verbose details? I couldn't think of one. And the question is whether that's really something vor the verbose output or should we not rather put it into the default output? The "resolved uri" (in terms of the hosts-array) is then only displayed in the verbose result, but the hosts_uri could be exposed on every call. Or am I completely wrong?

liam-wiltshire:
Maybe you're right - the other uris are non-verbose? Happy to go with what you think though!

heiglandreas:
There are some URIs in the verbose-section but comments, talks, tracks and images-URIs are in the non-verbose one. Only the all-talks URI is in the verbose one.

I'll add the URI to the non-verbose section. Stay tuned.

PR #371, opened 2016-10-28
Adds a way to add hosts to an event

heiglandreas:
This PR adds a way to add hosts to an event by providing an API-Endpoint for POST-requests to `events/[event-id]/hosts`

This PR depends on #370 and should be rebased once that PR is merged.


src/models/UserMapper.php
```diff
@@ -544,6 +544,31 @@ public function getUserIdFromUsername($username)
     }
 
     /**
+     * Function to get just the user ID
+     *
+     * @param string $fullname The users full name we're looking for
+     *
+     * @return $user_id The user's ID (or false, if we didn't find her)
+     */
+    public function getUserIdFromFullname($fullname)
```
liam-wiltshire: @heiglandreas What happens if we have two users with the same name?

heiglandreas: We don't ever have 2 users with the same name. Ever! 

Damn! Usually that's my part to spot it. No idea what what happened there. I'll check that back with the legacy site as I was under the impression that you could add a user with it's full name there and didn't further think about it. I'll probably change it to the users login-name.

Thanks for spotting!!

liam-wiltshire: @heiglandreas Did you manage to get to the bottom of this?

heiglandreas: It actually works that way that I can add a user as host by it's name. I'll still need to dig into the code-portion of that though. Tonight looks promising...

heiglandreas:
Rebased onto the current master and added some more tests (found a bug in my last PR :frowning_face: )

PR #372, opened 2016-10-28
Updating API to use proper models and collections

liam-wiltshire:
I was having some problems on the frontend due to the way I had built the API, so this re-factors the endpoint to use models and collections properly!


liam-wiltshire:
@railto, if you get 10 mins, could you look at merging this for me please? https://github.com/joindin/joindin-web2/pull/384 (which this depends on), is a blocker for deploying web2

PR #373, opened 2016-10-31
JOINDIN-717 #close

cal-tec:
If you comment on an event with a rating of 0 then the comment gets created successfully with an HTTP 201 but looking at the code the error that is displayed if a comment is rated > 5 is 'The field "rating" must be a number (1-5)'
This implies that a comment with a 0 rating should return the same error and not a 201 Created

Due to an irc discussion it became apparent that the constraints on the rating field should only be applied when a rating was allowable (if not previously rated and not an event host). It was however indicated that the rating field should be required in all situations to make the api easier to use & document.

This by default solves a fairly major (and i believe unidentified) bug in that if allowable a rating could be submitted of any value of 5 or less (including negative values).

### DIFF
--- a/src/controllers/Event_commentsController.php
+++ b/src/controllers/Event_commentsController.php
@@ -74,8 +74,8 @@ class Event_commentsController extends ApiController
         $rating = $request->getParameter('rating', false);
         if (false === $rating) {
             throw new Exception('The field "rating" is required', 400);
-        } elseif (false === is_numeric($rating) || $rating > 5) {
-            throw new Exception('The field "rating" must be a number (1-5)', 400);
+        } elseif (false === is_numeric($rating)) {
+            throw new Exception('The field "rating" must be a number', 400);
         }
 
         $commentText = $request->getParameter('comment');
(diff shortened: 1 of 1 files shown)


heiglandreas:
Just to clarify: 

> Due to an irc discussion it became apparent that the constraints on the rating field should only be applied when a rating was allowable (if not previously rated and not an **event host**). It was however indicated that the rating field should be required in all situations to make the api easier to use & document.

Should that not be the speaker of the talk and not the event host?

cal-tec:
No it is supposed to read like that, this is related to event comments only.

heiglandreas:
Sorry, TL/DR... YOu'Re absolutely right!

src/controllers/Event_commentsController.php
```diff
@@ -74,8 +74,8 @@ public function createComment($request, $db)
         $rating = $request->getParameter('rating', false);
         if (false === $rating) {
             throw new Exception('The field "rating" is required', 400);
-        } elseif (false === is_numeric($rating) || $rating > 5) {
-            throw new Exception('The field "rating" must be a number (1-5)', 400);
+        } elseif (false === is_numeric($rating)) {
```
heiglandreas: Can we remove the ```elseif``` here? There's an exception thrown before so the method is left there already. 

I'd prefer two separate ```if```-statements without the else.

PR #375, opened 2016-11-03
PoC for an event-based notification system for the joindin-api

heiglandreas:
This is currently based on the Symfony EventDispatcher but the backend can easily be swapped against something else.

Usage:

Event-Listeners (implementing the ```ListenerInterface``` are added to an ```EventCoordinator``` before the route is dispatched. 

That ```EventCoordinator``` is then injected into each route-method beneath the db-adapter and the request-object. And from there it can be injected into the Mapper-Objects.

In the controllers and the Mappers that ```EventCoordinator``` can then trigger an event with an event-Object as parameter. That event-Object needs to be specific for the Event and needs to extend the ```AbstractEvent```.

As example I removed the Email-Notification that is sent when a new Event is created from the ```EventController``` and added an ```EmailListener``` and a ```PendingEventCreated``` Notification to the  ```EventMapper```.

This is a base for discussion whether such a system makes sense for the joindin-API and if so, whether this is a possible way of doing so.

Thanks for your feedback!

(The dependency is currently only added to the composer.json, not yet to the vendor-folder, That should first happen when we want to do it this way!)


liam-wiltshire:
Thanks for putting this together @heiglandreas. I agree with the problem we are trying to solve (reducing the copy-paste code that comes with having lots of emails), however I'm not convinced this is the way to do it.

This could certainly do it, however I think (especially keeping in mind our target contributor audience), it's adding complexity that isn't needed.

I do think we could perhaps use the your https://github.com/heiglandreas/joindin-api/blob/741a410eb61974b6c2ab62de60fc30ada9ca11b2/src/services/EventbasedEmailService.php as the starting point for a genericised EmailService however - don't know if this is something you'd be happy to work on, or if you are ok to me to pinch your code and work on it when I get some time?

heiglandreas:
First things first: There's no problem when you want to pinch my code :)

But are you really sure such an event-based system is adding complexity? Yes, PubSub adds complexity on the architecture-level, agreed on that. But from what I've experienced so far it removes complexity on the Implementation-level. Why? Because it removes the need to think about too many things while implementing something. On implementing a controller you can stay focussed on what the controller should do. On Implementing a mapper you can focus on what the mapper should do. Yes, you have to remember to trigger an event. But in both cases you do not need to think about whether (and how) to implement possible notification or other features. And when you want to implement the notification you can concentrate on the notification without having to worry about where it should be implemented. So all it all I believe it will allow us to have cleaner code.

In my eyes that reduces the complexity on the implementation level pretty much. And from what I've seen so far thats the level a Junior developer will see on his first contribution. 

I don't mind throwing that code away and going without it. But I think it's a step towards a more complex application. 

When you are worried about the event-manager in the background: That was my idea of using existing components. But it would easily possible to write an event-manager  from scratch and reduce the dependencies of the project. That wouldn't bloat the code and would perhaps also be a way to show beginners how PubSub-systems work. 

What do others think?

heiglandreas:
Besides: I just had a look at the contributors list of the project: Either we need to redefine the target contributor audience or we need to attract some contributors from said audience… 😉

richsage:
We've had a few projects at work that have used an event-driven approach like this. Whilst as @heiglandreas says, it reduces controller complexity (which is good), it then spreads out your code into various places and it can, without good documentation, quickly become a sprawl of "what's listening to what?", "if this event fires, what actually happens?". Which then _increases_ the complexity outside of the controller itself.

Maybe it's just down to good documentation here, but I'd be worried about that complexity barrier if we were trying to keep the project open to all new contributors regardless of skill level :-)

heiglandreas:
So it doesn't increase *code complexity* but *documentation complexity*. And yes, it spreads your code into multiple places as it separates concerns. So the single pieces of *code* are smaller and easier to understand and maintain. 

I'm completely with you that the complexity rises the more gets triggered by events. That's one of the issues with ZF and Symfony. But contrary to those we don't have a DI-Container here that further "hides" the  dependencies. I was thinking about that for joindin as well but wouldn't want to introduce it as it would raise the complexity by hiding away dependencies. Events - especially when you have the Eventmanager (whatever that centtral piece of event-handling is called) injected into the relevant classes. That's the documentation of where it can be used. And the piece of code that injects the event listeners into that event-manager is the documentation of what might happen. 

So it's in a kind of way self-documented and there's no *magic* going on. 

Besides that, as it separates the concerns the single pieces of code are easier testable.

To make it easier to find EventListeners, I'd put them all into one namespace (```\Joindin\Api\EventListener``` f.e.) so they are not scattered around. And whether they are calling different classes or the controller itself doesn't make that much difference in added complexity.

On the contrary. As the dependencies for the EventListeners are (should be) given at the instantiation it is clear what the dependencies are against the current way of instantiating Services on the fly in the controller. 

So I would expect a cleaner and easier to handle codebase. Which in my eyes should make it easier for new (and old) contributors.

liam-wiltshire:
Hi Both,

I'm going to close this for now - I agree with the notion that something needs doing - my gut instinct is that a genericised Email service would be sufficient for our needs however.

Of course, this might change over time, and if enough people contact me and tell me I'm wrong, I'll probably change my mind :-p

PR #376, opened 2016-11-12
JOINDIN-642

exussum12:
Added ability to reject a talk claim, Updated the emails going out to
the user.

Also did some tidy up of code.

Updated Web2 not sure if legacy / the apps have the ability to reject a
talk so left the default behavour unchanged.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -750,4 +750,79 @@ class TalksController extends ApiController
         header('Location: ' . $uri, null, 204);
         exit;
     }
+
+    public function removeSpeakerForTalk(Request $request, PDO $db)
+    {
+        if (!isset($request->user_id)) {
+            throw new Exception("You must be logged in to delete data", 401);
+        }
+
+        $talk_id = $this->getItemId($request);
+
+        $talk_mapper = $this->getTalkMapper($db, $request);
+        $talk = $talk_mapper->getTalkById($talk_id);
+
+        $event_id = $talk->event_id;
+        $event_mapper = $this->getEventMapper($db, $request);
+        $event = $event_mapper->getEventById($event_id);
+
+        if (!$talk) {
+            throw new Exception("Talk not found", 404);
+        }
+
+        $is_admin = $talk_mapper->thisUserHasAdminOn($talk_id);
+        if (!($is_admin)) {
+            throw new Exception("You do not have permission to reject the speaker claim on this talk", 403);
+        }
+
+        $data = $this->getLinkUserDataFromRequest($request);
+
+        $user_mapper = $this->getUserMapper($db, $request);
+        $speaker_id = $user_mapper->getUserIdFromUsername($data['username']);
+        if (! $speaker_id) {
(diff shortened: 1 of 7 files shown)


heiglandreas:
Thanks for the great work! But I'd rather see that splitted into two PRs. One for the optimizations and one for the new feature.

Would that be possible?

exussum12:
There's only 1 thing which is not directly related to the pull ?  Removing an else statement. 

That whole function should be fixed at some point but didn't want to go too far down that path on an unrelated ticket

exussum12:
Added some comments

src/controllers/TalksController.php
```diff
@@ -562,7 +562,9 @@ public function setSpeakerForTalk(Request $request, PDO $db)
 
         if ($claim === false) {
             throw new Exception("No speaker matching that name found", 422);
-        } elseif ($claim['speaker_id'] != null && $claim['speaker_id'] != 0) {
+        }
+
+        if ($claim['speaker_id'] != null && $claim['speaker_id'] != 0) {
```
exussum12: Changed to remove an else, Functionally identical

tests/controllers/TalksControllerTest.php
```diff
@@ -644,7 +644,8 @@ public function testApproveClaimAsHostSucceeds()
         $request->user_id = 2;
         $request->parameters = [
             'username'      => 'psherman',
-            'display_name'  => 'P Sherman'
+            'display_name'  => 'P Sherman',
+            'action'        => 'approve',
```
exussum12: Not really needed the test works without this line too but this way is more correct

heiglandreas:
As far as I understood [JOINDIN-642](https://joindin.jira.com/browse/JOINDIN-642) the implementation should be a DELETE-Request to a pending claim. So removing a pending claim should be callable via  ``DELETE http://api.joind.in/v2.1/talks/[talk-id]/speakers/[speaker-id]```. That would mean a new route.

Appart from that there are some refactorings in this PR (https://github.com/joindin/joindin-api/pull/376#discussion-diff-87698178R647, https://github.com/joindin/joindin-api/pull/376/files#diff-ba9e65a421c84bdc956a3c7bca4d87a1R54 and https://github.com/joindin/joindin-api/pull/376/files#diff-ecab90309901903ed3917a59d9f52079R582 f.e.) that have nothing to do with the new functionality and that would be better off in their own PR. Having this PR depend on that function is no problem at all, but here we have two things mangled together that would be better of separated. Besides that it is easier to digest and therefore approve small separate PRs that one large one.

exussum12:
Updated the delete call

liam-wiltshire:
ok to test

liam-wiltshire:
Quite possibly the most minor violation possible, but if you could sort that line-length error so that Jenkins will build that'd be awesome :-)

exussum12:
on the webpr ?

exussum12:
Sorry thought I changed that, That section isnt needed any more - removed all of the code

PR #378, opened 2016-11-16
Fixes issues with newer phpunit-versions

heiglandreas:
```getMock``` is deprecated in newer phpunit-versions. THis PR replaces them with calls to ```getMockBuilder()```


exussum12:
Missed a few \Request replacements. Line 513 for example

heiglandreas:
You're right. Thanks for spotting. Those replacements shouldn't be part of this PR at all 😄

exussum12:
Thats a shame :( but :+1 for removing the warning :)

PR #379, opened 2016-11-16
Hotfix/header

heiglandreas:
This PR alters the way the HTTP-Status-code is set. It also replaces all the calls to ```header``` and ```exit``` with calls to ```ApiView::setHeader()```, ```ApiView::setResponseCode()```  and ```return```.

That makes way to easier testing.

### DIFF
--- a/src/controllers/EventImagesController.php
+++ b/src/controllers/EventImagesController.php
@@ -94,7 +94,12 @@ class EventImagesController extends ApiController
         $event_mapper->saveNewImage($event_id, $small_filename, $small_width, $small_height, "small");
 
         $location = $request->base . '/' . $request->version . '/events/' . $event_id;
-        header('Location: ' . $location, null, 201);
+
+        $view = $request->getView();
+        $view->setHeader('Location', $location);
+        $view->setResponseCode(201);
+        return;
+
     }
 
     public function deleteImage($request, $db)
@@ -108,7 +113,11 @@ class EventImagesController extends ApiController
         $event_mapper->removeImages($event_id);
 
         $location = $request->base . '/' . $request->version . '/events/' . $event_id;
-        header('Location: ' . $location, null, 204);
-        exit;
+
+        $view = $request->getView();
+        $view->setHeader('Location', $location);
+        $view->setResponseCode(204);
+        return;
+
     }
 }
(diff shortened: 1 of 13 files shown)


exussum12:
Great change ! Instead of just changing the tests to assertnull could they check the correct status code is being set the assert null test doesn't really gain anything

heiglandreas:
You're right @exussum12 in that they gain anything. But there wasn't anything to gain before as they simply asserted that the constructor was left before the headers were set. Therefore this PR just reinstates the current state. It's a refactoring that doesn't introduce new features.

But with this PR merged it's much easier to test controllers and a further PR can then add some more meaningful tests.

src/controllers/EventImagesController.php
```diff
@@ -94,7 +94,12 @@ public function createImage($request, $db)
         $event_mapper->saveNewImage($event_id, $small_filename, $small_width, $small_height, "small");
 
         $location = $request->base . '/' . $request->version . '/events/' . $event_id;
-        header('Location: ' . $location, null, 201);
+
+        $view = $request->getView();
+        $view->setHeader('Location', $location);
+        $view->setResponseCode(201);
+        return;
```
exussum12: These returns are not needed.

tests/views/ApiViewTest.php
```diff
@@ -0,0 +1,84 @@
...
+     * @preserveGlobalState disabled
+     */
+    public function testThatHeadersAreSet()
+    {
+        $view = new ApiView();
+
+        $view->setResponseCode(202);
+        header('Foo: bar', false, 201);
+
+        ob_start();
```
heiglandreas: Sadly not. As **this** tests exactly that part that outputs the view. And the output is done via ```echo``` so I need to get the output buffer here. This test only checks whether the correct header is returned as that was changed (and not tested before) in the refactoring.

A separate test that checks the correct string value is returned should be added with a separate PR.

liam-wiltshire:
@heiglandreas, there seems to be an issue here somewhere - https://github.com/joindin/joindin-web2/blob/master/app/src/Talk/TalkApi.php#L333 is expecting a 204 respons, however after applying this PR, it's returning a 302, which breaks the frontend....

PR #380, opened 2016-11-27
JOINDIN-600 Dont email author their own comments

exussum12:
This ensures that speakers do not get an email telling them they have
just replied.

The method is currently un tested, happy to add some but the function is
a mess too so would probably need refactoring. Another ticket ?

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -124,11 +124,17 @@ class TalksController extends ApiController
                         $comment    = $comment_mapper->getCommentById($new_id);
                         $speakers   = $talk_mapper->getSpeakerEmailsByTalkId($talk_id);
                         $recipients = array();
+                        $commentAuthorEmail = $comments[0]['email'];
                         foreach ($speakers as $person) {
+                            if ($commentAuthorEmail == $person['email']) {
+                                continue;
+                            }
                             $recipients[] = $person['email'];
                         }
-                        $emailService = new TalkCommentEmailService($this->config, $recipients, $talk, $comment);
-                        $emailService->sendEmail();
+                        if (count($recipients) > 0) {
+                            $emailService = new TalkCommentEmailService($this->config, $recipients, $talk, $comment);
+                            $emailService->sendEmail();
+                        }
                         $uri = $request->base . '/' . $request->version . '/talk_comments/' . $new_id;
                         header("Location: " . $uri, true, 201);
                         exit;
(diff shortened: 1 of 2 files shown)


src/controllers/TalksController.php
```diff
@@ -124,11 +124,17 @@ public function postAction($request, $db)
                         $comment    = $comment_mapper->getCommentById($new_id);
                         $speakers   = $talk_mapper->getSpeakerEmailsByTalkId($talk_id);
                         $recipients = array();
+                        $commentAuthorEmail = $comments[0]['email'];
```
heiglandreas: Shouldn't that be ```$comment[0]['email']```?

src/controllers/TalksController.php
```diff
@@ -124,11 +124,17 @@ public function postAction($request, $db)
...
                         foreach ($speakers as $person) {
+                            if ($commentAuthorEmail == $person['email']) {
+                                continue;
+                            }
                             $recipients[] = $person['email'];
                         }
-                        $emailService = new TalkCommentEmailService($this->config, $recipients, $talk, $comment);
-                        $emailService->sendEmail();
+                        if (count($recipients) > 0) {
+                            $emailService = new TalkCommentEmailService($this->config, $recipients, $talk, $comment);
```
heiglandreas: So we could remove the ```if``` condition here.

src/controllers/TalksController.php
exussum12: Oops yes, This has been updated

src/controllers/TalksController.php
exussum12: removed from here and added to the send email

exussum12: 2 mins, just checked further down stream, there is a foreach which will just not run if there are no emails

src/controllers/TalksController.php
```diff
@@ -124,9 +124,14 @@ public function postAction($request, $db)
                         $comment    = $comment_mapper->getCommentById($new_id);
                         $speakers   = $talk_mapper->getSpeakerEmailsByTalkId($talk_id);
                         $recipients = array();
+                        $commentAuthorEmail = $comment[0]['email'];
```
heiglandreas: Is this additional assignment necessary? As you're not calling a function there it should be possible to check against ```$comment[0]['email']``` on every foreach loop.

heiglandreas: I'd say it's more readable without the additional assignment 😄 

Directly in the foreach I find a variable that I need to check what it contains before I can decide what it does. With the direct comparison I'd see it's both an email. 

Additionally the new variable is only used in this one single line. For me that usually is a sign that the assignment can be omitted.

exussum12: without looking at the code around it theres no way to know what  `$comment[0]['email']` is. The variable makes that clear (and readable without having to lookup the meaning)
Happy to change if required

liam-wiltshire:
Closed as duplicate.

PR #381, opened 2016-11-28
Joindin 600 tested

exussum12:
This PR includes the other one, This one adds tests can be merged separate or together.

Added a few methods to make this testable.


liam-wiltshire:
As far as I can see, this unfortunately doesn't work?

I've done a bit of digging, and `$comment_mapper->getCommentById` returns an array, but the top-level index is 'comments':

`
array(2) {
  'comments' =>
  array(1) {
    [0] =>
    array(12) {
      'rating' =>
      string(1) "0"
      'comment' =>
      string(28) "asdasdadssdpofidspoifdssadas"
      'user_display_name' =>
      string(9) "Ima Admin"
      'username' =>
      string(8) "imaadmin"
      'talk_title' =>
      string(36) "Creating technical entities in Linux"
      'created_date' =>
      string(25) "2016-12-18T15:49:42+01:00"
      'uri' =>
      string(47) "http://api.dev.joind.in/v2.1/talk_comments/1006"
      'verbose_uri' =>
      string(59) "http://api.dev.joind.in/v2.1/talk_comments/1006?verbose=yes"
      'talk_uri' =>
      string(37) "http://api.dev.joind.in/v2.1/talks/89"
      'talk_comments_uri' =>
      string(46) "http://api.dev.joind.in/v2.1/talks/89/comments"
      'reported_uri' =>
      string(56) "http://api.dev.joind.in/v2.1/talk_comments/1006/reported"
      'user_uri' =>
      string(36) "http://api.dev.joind.in/v2.1/users/1"
    }
  }
  'meta' =>
  array(3) {
    'count' =>
    int(1)
    'total' =>
    string(1) "1"
    'this_page' =>
    string(64) "http://api.dev.joind.in/v2.1/talks/89/comments?resultsperpage=20"
  }
}
`

So it's probably just a case of updating https://github.com/exussum12/joindin-api/blob/06972fddfdf664622326a5fe28947d91653dd343/src/controllers/TalksController.php#L127 to check that array element?

Thanks!

exussum12:
Fixed, Sorry missed that. Should have ran the code outside of the tests to double check. Missed the call which adds that element in

liam-wiltshire:
ok to test

liam-wiltshire:
I've just looked at this again (after resolving the conflict), and it still doesn't seem to be working - I did some playing, and I don't think that the comments return an email address - and thinking about it, it shouldn't do, as that would potentially be exposed in the API which could then be scraped.

As such, I think this needs to be approached from a different way - checking the user ID rather than email address for example.

Let me know if you want to take a look at this, otherwise if you don't have the time right now then feel free to let me know and I'll put it on JIRA.

Thanks,

Liam

exussum12:
Sorry, Tested properly this time! I have fixed the merge conflict and changed to be the ids

PR #384, opened 2016-12-12
Implements rate-limiting

heiglandreas:
This adds a possibility to check rate-limits for either a user or an application. Currently the implementation doesn't actually limit as it returns true every time the hasValidRateLimit is called.

This might be necessary for https://github.com/joindin/joindin-api/pull/383

I'd much more like to use a PSR7-middleware for that but that needs much more adaptions so I went for the easy way…

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,9 @@
 {
+    "autoload" : {
+        "psr-4" : {
+            "Joindin\\Api\\" : "src/"
+        }
+    },
     "preferred-install": "dist",
     "require" : {
         "swiftmailer/swiftmailer": "v5.0.3",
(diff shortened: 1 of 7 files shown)


akrabat:
If you're going to rate limit, ensure the you put these headers in all responses:

* `X-RateLimit-Limit`
* ` X-RateLimit-Remaining`
* `X-RateLimit-Reset`

i.e. Copy GitHub - see https://developer.github.com/v3/#rate-limiting

heiglandreas:
After a short conversation with @lornajane I decided to close this PR. As long as there is no issue with traffic on the API we won't need rate-limiting. Should the need arise, we can always check this code or create new one.

PR #388, opened 2017-01-05
Only show tokens in trusted apps

heiglandreas:
This PR will

1. Identify a trusted app by checking whether the access-token was issued by an app that has the ```enable_password_grant```-flag in the database set
2. Only display tokens issued by apps that have the (newly added) ```can_be_revoked```-flag in the database set.

For that a new flag in the ```oauth_consumers```-table has been added via patch-68.sql

### DIFF
--- a/db/patch69.sql
+++ b/db/patch69.sql
@@ -1,0 +1,9 @@
+-- Add a new field `can_be_revoked` to the oauth_consumers table
+-- That field describes whether tokens from this consumer can be revoked by
+-- a user or not. Only tokens from consumers where that flag is set will be shown
+-- in the API endpoint /token/
+ALTER TABLE `oauth_consumers` ADD COLUMN can_be_revoked (tinyint(1) DEFAULT '1');
+
+UPDATE `oauth_consumers` SET can_be_revoked = 0 WHERE consumer_key = 'web2';
+
+INSERT INTO patch_history SET patch_number = 69;
--- a/src/controllers/TokenController.php
+++ b/src/controllers/TokenController.php
@@ -66,7 +66,7 @@ class TokenController extends ApiController
 
         $mapper = $this->getTokenMapper($db, $request);
 
-        $tokens = $mapper->getTokensForUser(
+        $tokens = $mapper->getRevokableTokensForUser(
             $request->user_id,
             $this->getResultsPerPage($request),
             $this->getStart($request)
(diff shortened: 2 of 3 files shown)


railto:
Had a look through and seen that there is at least 1 more PR open at present #390 that has a patch68.sql, may need to coordinate with @exussum12 regarding numbering

heiglandreas:
The numbering of the patches needs to follow the sequence of merging. In the end the first PR merged wins as the second one with a patch with the same number will then contain a merge-conflict that needs to be resolved then by renumbering the patch. 

So in my eyes it's not only OK that both have the same patch number, it's even required to be able to choose which PR to merge first.

railto:
No worries @heiglandreas that works for me.  My comment yesterday was slightly incorrect, I meant to say there was a merge conflict on the file.  Just double checked master branch and can see we already have a patch68.sql.  Appears to be conflicting with the patch for updating the charset so this would need to be renamed to patch69.sql in order for it to be pulled in.

heiglandreas:
So you wanted to slightly hint to me that I can't get my own PRs in order? 😂

Got the hint then 😊

db/patch68.sql
```diff
@@ -3,4 +3,4 @@
 -- This resolves https://joindin.jira.com/browse/JOINDIN-725
 ALTER DATABASE joindin CHARACTER SET utf8 COLLATE utf8_unicode_ci;
 
-INSERT INTO patch_history SET patch_number = 68;
\ No newline at end of file
+INSERT INTO patch_history SET patch_number = 69;
```
railto: Maybe I'm being really stupid here (possible), but should this one not be left as patch_number = 68 for this file?

railto: To further this, just checked and can see that patch_level 68 is not being set in the DB, likely as there are 2 patches declaring as 69.

https://www.dropbox.com/s/5ic7qwkza45s3h7/Screenshot%202017-02-02%2013.39.23.png?dl=0

exussum12: Would argue that these should just be ticket-ref.sql and the person who merges renames and adds this line. Asking for conflicts the way its currently set up

exussum12: The number is only important once merged, not knowing before hand what order they will be merged makes the job impossible

railto: I would disagree there @exussum12, the person accepting the the merge should not be making any changes at all to the code, this should be done by the person that creates the change in the first place.

exussum12: That file has no meaning at all though. The code changes I agree, but this needs to be run on the DB.When its been run a simple rename (no PR needed) and insering the line to keep the dev DB up to date would be fine. The current method is set up for 1 PR at a time (which is less than ideal for growth)

heiglandreas: That was nothing but a rebase gone south… Fixing that ASAP

heiglandreas:
@railto Sorry for the inconvenience. Fixed the issue with the Patch-Numbering.

PR #389, opened 2017-01-05
OAuthModel throws exception on non-verified user login

railto:
This is part of a wider update for #JOINDIN-718 to show a user a link to the resend verification page if they try to login when not verified.

### DIFF
--- a/src/models/OAuthModel.php
+++ b/src/models/OAuthModel.php
@@ -125,16 +125,22 @@ class OAuthModel
      */
     protected function getUserId($username, $password)
     {
-        $sql  = 'SELECT ID, password, email FROM user
-            WHERE username=:username
-            AND verified = 1';
+        $sql  = 'SELECT ID, password, email, verified FROM user
+            WHERE username=:username';
         $stmt = $this->_db->prepare($sql);
         $stmt->execute(array("username" => $username));
         $result = $stmt->fetch(PDO::FETCH_ASSOC);
+
         if ($result) {
-            if (password_verify(md5($password), $result['password'])) {
-                return $result['ID'];
+            if ($result['verified'] == 1) {
+                if (password_verify(md5($password), $result['password'])) {
+                    return $result['ID'];
+                }
+
+                return false;
             }
+
+            throw new Exception("Not Verified", 401);
         }
 
         return false;
--- a/src/models/OAuthModel.php
+++ b/src/models/OAuthModel.php
@@ -131,20 +131,17 @@ class OAuthModel
         $stmt->execute(array("username" => $username));
         $result = $stmt->fetch(PDO::FETCH_ASSOC);
 
-        if ($result) {
-            if ($result['verified'] == 1) {
-                if (password_verify(md5($password), $result['password'])) {
-                    return $result['ID'];
-                }
+        if (!$result) return false;
 
-                return false;
-            }
-
-            throw new Exception("Not Verified", 401);
+        if ($result['verified'] != 1) {
+            throw new Exception("Not verified", 401);
         }
 
-        return false;
+        if (!password_verify(md5($password), $result['password'])) {
+            return false;
+        }
 
+        return $result['ID'];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/models/OAuthModel.php
```diff
@@ -125,16 +125,22 @@ public function createAccessTokenFromPassword($clientId, $username, $password)
...
-            WHERE username=:username
-            AND verified = 1';
+        $sql  = 'SELECT ID, password, email, verified FROM user
+            WHERE username=:username';
         $stmt = $this->_db->prepare($sql);
         $stmt->execute(array("username" => $username));
         $result = $stmt->fetch(PDO::FETCH_ASSOC);
+
         if ($result) {
-            if (password_verify(md5($password), $result['password'])) {
```
heiglandreas: While you're at it, would it make sense to fail early and rewrite that whole conditional tree?

```php
if (! $result) return false;
if ($result['verified'] !== 1) return false;
// etc.
```

railto: So, like
`if (!$result) return false;

        if ($result['verified'] !== 1) {
            throw new Exception("Not Verified", 401);
        }

        if (password_verify(md5($password), $result['password'])) {
            return $result['ID'];
        }

        return false;`
?

heiglandreas: Yeah. Along that way.

I'd even say ```if (! password_verify …  return false;``` (or ```throw Exception```?

I always try to return early in error conditions and have the "expected" outcome right at the end of the function.

railto: Ahh yeah, not a bad idea.

heiglandreas: Pure laziness 😉 

 I hate to search nested ifs to find the "main path" of a function 😄

railto: Tbh, I never actually thought about it that way, and it makes perfect sense.  When it comes to programming, laziness is not 'always' a bad thing

src/models/OAuthModel.php
```diff
@@ -125,20 +125,25 @@ public function createAccessTokenFromPassword($clientId, $username, $password)
      */
     protected function getUserId($username, $password)
     {
-        $sql  = 'SELECT ID, password, email FROM user
-            WHERE username=:username
-            AND verified = 1';
+        $sql  = 'SELECT ID, password, email, verified FROM user
+            WHERE username=:username';
         $stmt = $this->_db->prepare($sql);
         $stmt->execute(array("username" => $username));
```
heiglandreas: Should this ```execute``` also be wrapped inside an ```if``` to handle failure?

railto: Not sure I follow why we would need to in this instance...

heiglandreas: well, things can go wrong and ```execute``` can return ```false``` instead of true… 

But I'm perfectly fine with not doing it!

railto: The if (!$result) should catch in that instance.  It's also not something that we appear to be doing in other places, tho I'm happy to put it in if it's required.

heiglandreas: You're absolutely right! The ```if (! $result)``` should take care of that! Thanks for the eye-opener 👍

src/models/OAuthModel.php
```diff
@@ -125,20 +125,25 @@ public function createAccessTokenFromPassword($clientId, $username, $password)
      */
     protected function getUserId($username, $password)
```
exussum12: No Test ?

railto: Good point, just trying to figure out how to actually test this functionality

railto: Working on a controller test for this, however for some weird reason the PDO mock isn't being passed from the TokenController to the OAuthModel, looking into it to see what I can do

liam-wiltshire:
@railto - are you still looking at writing a test for this?

src/models/OAuthModel.php
exussum12: Happy to help with writing the test here

railto:
Per IRC chat,  was having issues getting the PDO mock passed from the TokenController to OAuthModel and I don't have a lot of time at present to dig further into it.

railto:
Closing as this is now included in #397

PR #390, opened 2017-01-14
JOINDIN-648 allow different talk types

exussum12:
Added new backend access for different talk type medias

This adds the new data to the verbose getTalk call and also added a new endpoint for getting all of the media types.

/v2.1/talks/3/links

and 
/v2.1/talks/3?verbose=yes

Have changed.

There is a backwards compatible change for for slide links (when they get moved over with another ticket)

### DIFF
--- a/db/patch68.sql
+++ b/db/patch68.sql
@@ -1,0 +1,26 @@
+CREATE TABLE `talk_types_link` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `talk_id` int(11) NOT NULL,
+  `talk_type` int(11) NOT NULL,
+  `url` text NOT NULL,
+  `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
+  KEY `talk_id` (`talk_id`),
+  KEY `talk_type` (`talk_type`),
+  CONSTRAINT `talk_types_link_ibfk_1` FOREIGN KEY (`talk_id`) REFERENCES `talks` (`ID`),
+  CONSTRAINT `talk_types_link_ibfk_2` FOREIGN KEY (`talk_type`) REFERENCES `talk_types` (`ID`)
+) ENGINE=InnoDB;
+
+CREATE TABLE `talk_types` (
+  `ID` int(11) NOT NULL AUTO_INCREMENT,
+  `type` varchar(255) NOT NULL,
+  `display_name` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`ID`)
+) ENGINE=InnoDB;
+
+
+INSERT INTO `talk_types`(`ID`,`type`,`display_name`) VALUES
+(1,'Slides','slides_link'),
+(2,'Video','video_link'),
+(3,'Audio','audio_link');
+
--- a/db/patch68.sql
+++ b/db/patch68.sql
@@ -1,26 +1,32 @@
-CREATE TABLE `talk_types_link` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
-  `talk_id` int(11) NOT NULL,
-  `talk_type` int(11) NOT NULL,
-  `url` text NOT NULL,
-  `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-  PRIMARY KEY (`id`),
-  KEY `talk_id` (`talk_id`),
-  KEY `talk_type` (`talk_type`),
-  CONSTRAINT `talk_types_link_ibfk_1` FOREIGN KEY (`talk_id`) REFERENCES `talks` (`ID`),
-  CONSTRAINT `talk_types_link_ibfk_2` FOREIGN KEY (`talk_type`) REFERENCES `talk_types` (`ID`)
-) ENGINE=InnoDB;
+#Theres already an unsed table relating to this, drop it
+DROP TABLE talk_link_types;
+#switch talks to innodb so FKs work
+ALTER TABLE `joindin`.`talks` ENGINE=INNODB;
 
-CREATE TABLE `talk_types` (
-  `ID` int(11) NOT NULL AUTO_INCREMENT,
-  `type` varchar(255) NOT NULL,
-  `display_name` varchar(255) DEFAULT NULL,
+CREATE TABLE `talk_link_types` (
+  `ID` INT(11) NOT NULL AUTO_INCREMENT,
+  `display_name` VARCHAR(255) DEFAULT NULL,
   PRIMARY KEY (`ID`)
-) ENGINE=InnoDB;
+) ENGINE=INNODB;
 
 
-INSERT INTO `talk_types`(`ID`,`type`,`display_name`) VALUES
-(1,'Slides','slides_link'),
-(2,'Video','video_link'),
-(3,'Audio','audio_link');
+CREATE TABLE `joindin`.`talk_links`(
+  `ID` INT NOT NULL AUTO_INCREMENT,
+  `talk_id` INT NOT NULL,
+  `talk_type` INT NOT NULL,
+  `url` TEXT NOT NULL,
+  `timestamp` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  PRIMARY KEY (`ID`),
+  FOREIGN KEY (`talk_id`) REFERENCES `joindin`.`talks`(`ID`),
+  FOREIGN KEY (`talk_type`) REFERENCES `joindin`.`talk_link_types`(`ID`)
+);
 
+
+INSERT INTO `talk_link_types`(`ID`,`type`,`display_name`) VALUES
+(1,'slides_link'),
+(2,'video_link'),
+(3,'audio_link'),
+(4,'code_link'), #github / bitbucket links
+(5,'joindin_link'); #Links to another talk on the same topic by the speaker
+
+INSERT INTO patch_history SET patch_number = 67;
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -27,7 +27,7 @@ class TalksController extends ApiController
             }
         } else {
             if ($talk_id) {
-                $talk = $this->getTalkById($db, $request, $talk_id);
+                $talk = $this->getTalkById($request, $db, $talk_id, $verbose);
                 $collection = new TalkModelCollection([$talk], 1);
                 $list = $collection->getOutputView($request, $verbose);
             } elseif (isset($request->parameters['title'])) {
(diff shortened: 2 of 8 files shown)


db/patch68.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE `talk_types_link` (
```
exussum12: Not sure if this is the correct way to do DB changes ? Its basically these two tables

src/controllers/TalksController.php
```diff
@@ -51,13 +51,11 @@ public function getAction($request, $db)
 
     public function postAction($request, $db)
     {
-        if (! isset($request->user_id)) {
-            throw new Exception("You must be logged in to create data", 401);
-        }
+        $this->checkLoggedIn($request);
```
exussum12: Refacored all of these, They are everywhere.

src/controllers/TalksController.php
```diff
@@ -51,13 +51,11 @@ public function getAction($request, $db)
...
         $talk_id = $this->getItemId($request);
 
         // Retrieve the talk. It if doesn't exist, then 404 with talk not found
-        $talk= $this->getTalkById($db, $request, $talk_id);
+        $talk = $this->getTalkById($request, $db, $talk_id);
```
exussum12: Argument order switched for consistency

src/controllers/TalksController.php
```diff
@@ -215,16 +211,11 @@ public function deleteAction($request, $db)
      */
     public function addTrackToTalk(Request $request, PDO $db)
     {
-        if (!isset($request->user_id)) {
-            throw new Exception("You must be logged in to create data", 400);
```
exussum12: Oddly a few of the existing requests had 400 response for this ? I had changed all of these to 401. Possible breaking change

src/controllers/TalksController.php
```diff
@@ -822,16 +788,46 @@ public function removeSpeakerForTalk(Request $request, PDO $db)
             throw new Exception("There was a problem assigning the talk", 500);
         }
 
-        //If we are unit testing, then we can't exit or send headers!
-        if (defined('UNIT_TEST')) {
```
exussum12: adjusted these to match @heiglandreas  change. His was merged while this pull request was open.

src/models/OAuthModel.php
```diff
@@ -167,11 +167,9 @@ public function newAccessToken($consumer_key, $user_id)
     {
         $hash              = $this->generateToken();
         $accessToken       = substr($hash, 0, 16);
-        $accessTokenSecret = substr($hash, 16, 16);
 
         $sql = "INSERT INTO oauth_access_tokens set
                 access_token = :access_token,
-                access_token_secret = :access_token_secret,
```
exussum12: This was never used, Removed

tests/controllers/TalksControllerTest.php
```diff
@@ -18,10 +23,16 @@ class TalksControllerTest extends \PHPUnit_Framework_TestCase
      */
     public function testClaimTalkWithNoUserIdThrowsException()
     {
-        $request = new \Request([], ['REQUEST_URI' => "http://api.dev.joind.in/v2.1/talks/326/speakers", 'REQUEST_METHOD' => 'POST']);
+        $request = new Request(
```
exussum12: These were all > 120. Reduced to < 120 for PSR-2

tests/controllers/TalksControllerTest.php
```diff
@@ -33,17 +44,23 @@ public function testClaimTalkWithNoUserIdThrowsException()
      *
      * @return void
      *
-     * @test
```
exussum12: Removed a few of these too - all of the functions are prefixed with test so seems pointless to have the test annotation too

db/patch68.sql
heiglandreas: Are we talking about talk_types or slides?

db/patch68.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE TABLE `talk_types` (
+  `ID` int(11) NOT NULL AUTO_INCREMENT,
+  `type` varchar(255) NOT NULL,
+  `display_name` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`ID`)
+) ENGINE=InnoDB;
+
+
+INSERT INTO `talk_types`(`ID`,`type`,`display_name`) VALUES
+(1,'Slides','slides_link'),
```
heiglandreas: Shouldn't that be (1, 'slides_link', 'Slides')?

db/patch68.sql
```diff
@@ -0,0 +1,26 @@
...
+(2,'Video','video_link'),
+(3,'Audio','audio_link');
+
```
heiglandreas: At the end you should have a line ```INSERT INTO patch_history SET patch_number = 68;```

src/controllers/TalksController.php
heiglandreas: I'd prefer to have those refactorings in a separate PR. They aren't necessary for the feature. Having them in two separate PRs makes each one easier to digest :)

db/patch68.sql
exussum12: Slides is a form of talk type. Though agreed the name could be better. Any suggestions ?

db/patch68.sql
exussum12: display_name here is the display name when looking at the endpoint. Its currently slides_link and should not really change (as the front end would need some additional code to handle it)

db/patch68.sql
exussum12: Will add that thanks :)

db/patch68.sql
heiglandreas: Aren't they talk_links?

src/controllers/TalksController.php
heiglandreas: We had a discussion about that a while back and decided to leave them at the moment as they are. If changed that should also be a separate PR that then changes *all* of those occurences throughout the API

src/controllers/TalksController.php
heiglandreas: This also (despite being a good one) should also be a separate PR.

src/models/OAuthModel.php
heiglandreas: Is that necessary for the talk links? Otherwise please add that as a further PR. 

And even though it's not used ATM it might become necessary in the future.

tests/controllers/TalksControllerTest.php
heiglandreas: Also please add this as a separate PR.

tests/controllers/TalksControllerTest.php
heiglandreas: Though you are absolutely right, that's a refactoring that should also go into it's own PR

src/controllers/TalksController.php
```diff
@@ -215,16 +211,11 @@ public function deleteAction($request, $db)
...
-        if (!isset($request->user_id)) {
-            throw new Exception("You must be logged in to create data", 400);
-        }
+        $this->checkLoggedIn($request);
 
-        $talk_id = $this->getItemId($request);
         $talk_mapper = new TalkMapper($db, $request);
-        $talk = $talk_mapper->getTalkById($talk_id);
-        if (!$talk) {
-            throw new Exception("Talk not found", 404);
```
heiglandreas: Why do you need to remove this line? Is that relevant to the talk-links?

Removing it has two downsides: 

1. line 218 might fail when the user asks for a talk not available
2. The user will never know that the talk doesn't exist.

src/controllers/TalksController.php
```diff
@@ -268,18 +259,13 @@ public function addTrackToTalk(Request $request, PDO $db)
...
-        }
+        $this->checkLoggedIn($request);
 
-        $talk_id = $this->getItemId($request);
         $track_id = $request->url_elements[5];
 
         $talk_mapper = new TalkMapper($db, $request);
-        $talk = $talk_mapper->getTalkById($talk_id);
-        if (!$talk) {
-            throw new Exception("Talk not found", 404);
```
heiglandreas: see above…

src/models/OAuthModel.php
exussum12: Then add it in the future when it *is* needed. Keeping dead code in the codebase is silly

heiglandreas: It's already there. And removing it is not the purpose of this PR.

src/controllers/TalksController.php
```diff
@@ -320,10 +306,18 @@ public function removeTrackFromTalk(Request $request, PDO $db)
      *
      * @return TalkModelCollection
      */
-    protected function getTalkById($db, $request, $talk_id)
-    {
-        $mapper = new TalkMapper($db, $request);
-        $talk   = $mapper->getTalkById($talk_id);
+    protected function getTalkById(
```
heiglandreas: Rewriting this method is not necessary for the scope of the PR. so please add it as a separate PR.

db/patch68.sql
heiglandreas: Then what is the type needed for? A display name in my eyes is something that should be displayed to an end user. Have a look at https://github.com/joindin/joindin-api/blob/master/src/controllers/LanguagesController.php for an example

src/models/OAuthModel.php
```diff
@@ -181,7 +179,6 @@ public function newAccessToken($consumer_key, $user_id)
         $result = $stmt->execute(
             array(
                 'access_token'        => $accessToken,
-                'access_token_secret' => $accessTokenSecret,
```
heiglandreas: See above.

tests/controllers/TalksControllerTest.php
```diff
@@ -775,6 +861,50 @@ public function testRejectClaimAsHostSucceeds()
...
+            ]
+        );
+        $request->user_id = 1;
+        $request->parameters = [
+            'username'      => 'psherman',
+            'display_name'  => 'P Sherman',
+            'verbose'       => 'yes',
+        ];
+        $db = $this->getMockBuilder(mockPDO::class)->getMock();
+        $stmt = $this->getMockBuilder(PDOStatement::class)
```
heiglandreas: It would be better to inject a mocked TalkMediaMapper into the controller. That way you can test the controller independent from the mapper. That would then also require tests for the TalkMediaMapper.

db/patch68.sql
exussum12: Should an API have any say at all on the front end appearance ?

src/controllers/TalksController.php
exussum12: It is needed for the verbose element to be added

tests/controllers/TalksControllerTest.php
exussum12: Good shout. Was testing too much. I have moved that now

db/patch68.sql
exussum12: added thanks

src/controllers/TalksController.php
exussum12: Changed all of them back to throwing 400's (which did previously)

db/patch68.sql
exussum12: Changed the names.

exussum12:
Not sure if we want to just move this to the other PR ? Conflict resolved and as it built on this contains both

exussum12:
Closing due to all of this being contained in the other PR

PR #392, opened 2017-01-24
JOINDIN-649 Add in CRUD for talk links

exussum12:
Delete, Post, Put, Get added
Changed the structure of the test to make smaller tests

Had to adjust the autoloader because of this

All lines changes are covered by tests

PR #393, opened 2017-01-24
JOINDIN-651 Back populate new table

exussum12:
Probably need a discussion on how to backpopulate the internal links.

Speaker + talk title I think would be a match ? Any other suggestions ?

### DIFF
--- a/db/patch71.sql
+++ b/db/patch71.sql
@@ -5,6 +5,6 @@ SELECT
   slides_link
 FROM
   talks
-WHERE slides_link <> ''
+WHERE slides_link <> '';
 
 INSERT INTO patch_history SET patch_number = 71;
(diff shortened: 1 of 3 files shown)


railto:
ok to test

railto:
This can not be merged until #392 is first merged, note the patch number will need to be increased to 71

exussum12:
@liam-wiltshire just a suggestion with these patch numbers, can they be ticket numbers until they are merged ? They only mean something when they are merged and its impossible to know before hand which numbers will be free

liam-wiltshire:
You ok to update this to 71 please? :)

db/patch71.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO `joindin`.`talk_links` (`talk_id`, `talk_type`, `url`)
+SELECT
+  id,
+  1,
+  slides_link
+FROM
+  talks
+WHERE slides_link <> ''
```
liam-wiltshire: This is currently throwing a syntax error in mySQL - think it's just a missing semi-colon however.

exussum12: Your right. Will fix now sorry

PR #396, opened 2017-02-17
Fix: Do not check in vendor directory

localheinz:
This PR

* [x] adds `vendor` to `.gitignore` and removes checked in dependencies

💁‍♂️ Not sure, is there a reason for them to be checked in? That's what we use `composer` for, right?


railto:
ok to test

akrabat:
`vendor` is checked in because the server does not have composer installed on it and the deployment scripts do not handle it.

When we set up to use composer, it did not do package signing or checksumming and we didn't trust that we couldn't guarantee that the code deployed via a script would be exactly what we tested against. I have no idea if this has changed.

railto:
@localheinz thank you for the contribution. Accepting this will require some changes on the build test server as well as the test/production servers.  This will need to be discussed with @liam-wiltshire and others before proceeding.

heiglandreas:
There was a reason tho have them checked in! It makes the deployment-process independent from packagist etc.

Though that is a decission from the starting days of joind.in I wouldn't want this to happen without a discussion on IRC before as in my eyes it's an architectural decission.

akrabat:
FWiW, https://github.com/composer/composer/issues/1074 was only fixed last year and that was certainly one of the concerns when we first started using composer with joind.in.

liam-wiltshire:
For me, simply the fact that we might deploy to test but then not deploy to production for a little while justifies having it checked in. As much as revision changes shouldn't break things, I've still known them to!

This comes up every now and again, and the concensus is to stick with the status quo generally. Happy to discuss it further in IRC if anyone strongly objects, but as it stands I'd personally leave it as it is for now.

localheinz:
Ha, interesting - how about using `composer` then as part of the deploy pipeline?

PR #398, opened 2017-02-18
JOINDIN-730 Clicking claim twice

exussum12:
This will be a breaking change if anywhere else approves claims apart
from web2.

The button on the talk page calls the same end point with the same data
as the admin one.

The count of the call is the differenciator.

This PR adds some data in the call to know which place you came from

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -584,10 +584,12 @@ class TalksController extends BaseTalkController
             } else {
                 throw new Exception("You must be the speaker or event admin to link a user to a talk", 401);
             }
-        } elseif ($claim_exists === PendingTalkClaimMapper::SPEAKER_CLAIM) {
+        } elseif ((
+            $claim_exists === PendingTalkClaimMapper::SPEAKER_CLAIM &&
+            $this->getRequestParameter($request, 'action') == 'approve'
+        )) {
             //The host needs to approve
             if ($talk_mapper->thisUserHasAdminOn($talk_id)) {
-                $method = $this->getRequestParameter($request, 'action', 'approve');
                 $recipients   = [$user_mapper->getEmailByUserId($speaker_id)];
 
                 $success = $pending_talk_claim_mapper->approveClaimAsHost($talk_id, $speaker_id, $claim['ID'])
--- a/tests/controllers/TalksControllerTest.php
+++ b/tests/controllers/TalksControllerTest.php
@@ -536,7 +536,8 @@ class TalksControllerTest extends TalkBase
         $request->user_id = 2;
         $request->parameters = [
             'username'      => 'janebloggs',
-            'display_name'  => 'Jane Bloggs'
+            'display_name'  => 'Jane Bloggs',
+            'action'        => 'approve',
         ];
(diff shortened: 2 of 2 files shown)


railto:
ok to test

liam-wiltshire:
This is an issue, however I'm trying to think if we can avoid including an action in this way...

Basically, the reason why it's a shared endpoint is that eventually a talk claim will work from either end, so either the speaker can claim it, or the host can 'assign' it.

As such, I wonder if a better solution would be to simply change the error message? As far as I can see, that specific error is only shown if the claim already exists (So the same talk ID and the same speaker ID), and it has already been approved by the speaker. As such, I think just changing that error to "You have already submitted a claim for this talk, and we are waiting for the host to approve it. You will receive an email once this has been completed." or words to that effect would achieve the same thing?

exussum12:
If you have admin privileges. Pressing that button twice( even a double click) will assign you to the talk. 

They are distinct actions (claim and approve) and really should be two end points. Making the end point do two things violates the single responsibility principal. 

We could change it so an admin can't assign their own talk ?

liam-wiltshire:
Re Single responsibility, yes you are right - I can't remember now why we did it that way - I imagine it'll be in JIRA somewhere. We could split them up, but at the moment there are probably 2 or 3 things we need to do in order to essentially remove reliance on legacy, so my main concern is sorting those out first :).

I guess it comes down to what we see as the issue - some events only have one host, so if they were speaking they'd need to be able to approve their own talk. Likewise with site admins, the potential for a mis-click is there sure, but worst case they can always unlink it again.

As such, in my mind the main issue here is the messaging that comes back, which can be sorted fairly simply.

Happy to be outvoted, just trying to think of the simplest solution.

exussum12:
How about chnaging both of the web 2 calls to use 2 new end points. And leave the current one there (view a view of removing later ) 

Then nothing breaking backwards compatibility and the fix going forward is separated to two end points (a claim and an arrprove end point ) allowing an admin user to approve their own but on two pages ?

exussum12:
Or just another thought. Add a new end point / message for the claiming talk. If nothing is posted then default to old behaviour ? 

That way backwards compatibility isn't broken. It's a simple change to make (just applying to fix here to the other half of the call )

Thoughts ?

liam-wiltshire:
For reference, here are the original JIRA issues:

https://joindin.jira.com/browse/JOINDIN-639
https://joindin.jira.com/browse/JOINDIN-640

I think that the reason behind it being one endpoint was that was the best way to RESTfully represent it - as you are only really looking at one entity (a talk speaker), but that can either be claimed by a speaker, or assigned by a host, and then approved by whoever the other party is.

I have no issue with it being re-factored in some way, as long as:

- It keeps it's RESTfulness (and thus any related URIs in other APIs need to accurately reflect this if it does change)
- It doesn't cause issues in making it bi-directional down the line.

Then I'm still happy :)

liam-wiltshire:
Further to the above, I'm going to close this for now, however if someone want's to have a go at re-factoring it, then I'm more than happy to look at that :)

PR #399, opened 2017-02-19
Joindin 700

exussum12:
From 

image

To

image

First commit adds tests, second commit refactors. Split so both can be looked at, tests pass on both commits

### DIFF
--- a/src/controllers/BaseTalkController.php
+++ b/src/controllers/BaseTalkController.php
@@ -2,6 +2,18 @@
 
 class BaseTalkController extends ApiController
 {
+    /** @var PDO */
+    protected $db;
+    /** @var Request */
+    protected $request;
+
+    protected $classMappings = [
+        'talk' => TalkMapper::class,
+        'talkcomment' => TalkCommentMapper::class,
+    ];
+
+    protected $classMaps = [];
+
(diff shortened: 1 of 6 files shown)


src/controllers/BaseTalkController.php
```diff
@@ -65,19 +77,6 @@ public function getUserMapper($db, $request)
         return $this->user_mapper;
     }
 
-    public function setTalkCommentMapper(TalkCommentMapper $talk_comment_mapper)
-    {
-        $this->talk_comment_mapper = $talk_comment_mapper;
-    }
-
-    public function getTalkCommentMapper($db, $request)
```
exussum12: Going to get rid of most of these functions, Lots of copy/paste to make them

tests/controllers/TalkLinkControllerTest.php
```diff
@@ -3,12 +3,7 @@
 namespace JoindinTest\Controller;
 
 use JoindinTest\Inc\mockPDO;
-use PDOStatement;
```
exussum12: Just clean up, these dont appear to be used

exussum12:
Removed all the ifs by adding more routes, Should work as before

liam-wiltshire:
ok to test

PR #400, opened 2017-02-19
Fix: Remove unnecessary parentheses

localheinz:
This PR

* [x] removes unnecessary parentheses when using `include` or `require`

💁‍♂️ For reference, see 

* http://php.net/manual/en/function.include.php
* http://php.net/manual/en/function.require.php.


liam-wiltshire:
ok to test

PR #401, opened 2017-02-19
Fix: Remove unused parameters from method calls

localheinz:
This PR

* [x] removes unused parameters from method calls

### DIFF
--- a/src/controllers/EventsController.php
+++ b/src/controllers/EventsController.php
@@ -30,14 +30,12 @@ class EventsController extends ApiController
                     );
                     break;
                 case 'talk_comments':
-                    $sort                = $this->getSort($request);
                     $talk_comment_mapper = new TalkCommentMapper($db, $request);
                     $list                = $talk_comment_mapper->getCommentsByEventId(
                         $event_id,
                         $resultsperpage,
                         $start,
-                        $verbose,
-                        $sort
+                        $verbose
                     );
                     break;
                 case 'attendees':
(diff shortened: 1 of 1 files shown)


src/controllers/EventsController.php
```diff
@@ -30,14 +30,12 @@ public function getAction($request, $db)
                     );
                     break;
                 case 'talk_comments':
-                    $sort                = $this->getSort($request);
                     $talk_comment_mapper = new TalkCommentMapper($db, $request);
                     $list                = $talk_comment_mapper->getCommentsByEventId(
```
localheinz: The method [`TalkCommentMapper::getCommentsByEventId()`](https://github.com/joindin/joindin-api/blob/dd87d0258a8e0a0dec41da585f6574034295ccef/src/models/TalkCommentMapper.php#L56-L79) accepts only 4 parameters; `$sort` is never used.

src/controllers/EventsController.php
```diff
@@ -353,7 +351,7 @@ function ($tag) {
 
                 // Send an email if we didn't auto-approve
                 if (! $user_mapper->isSiteAdmin($request->user_id)) {
-                    $event        = $event_mapper->getPendingEventById($event_id, true);
+                    $event        = $event_mapper->getPendingEventById($event_id);
```
localheinz: The method [`EventMapper:: getPendingEventById()`](https://github.com/joindin/joindin-api/blob/436971cedc3df8708e0e53e617ae99eb5b198a92/src/models/EventMapper.php#L1006-L1022) accepts only one parameter.

liam-wiltshire:
ok to test

PR #402, opened 2017-02-19
Fix: Rename method to match invocation

localheinz:
This PR

* [x] fixes a case-insensitive method call

### DIFF
--- a/src/models/EventHostMapper.php
+++ b/src/models/EventHostMapper.php
@@ -56,7 +56,7 @@ class EventHostMapper extends ApiMapper
         return $this->_db->lastInsertId();
     }
 
-    public function removeHostfromEvent($host_id, $event_id)
+    public function removeHostFromEvent($host_id, $event_id)
     {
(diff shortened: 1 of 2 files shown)


src/controllers/Event_hostsController.php
```diff
@@ -141,7 +141,7 @@ public function removeHostFromEvent(Request $request, PDO $db)
 
         $mapper = $this->getEventHostMapper($request, $db);
 
-        $uid = $mapper->removeHostFromEvent($user_id, $event_id);
+        $uid = $mapper->removeHostfromEvent($user_id, $event_id);
```
localheinz: See [`EventHostMapper::removeHostfromEvent()`](https://github.com/joindin/joindin-api/blob/0e120fc5ea508e9483a926a1c82c91cb928f56ec/src/models/EventHostMapper.php#L59-L69).

Alternatively, of course, we can rename the method (which in my opinion makes more sense),

exussum12: Rename the function please

liam-wiltshire:
ok to test

PR #407, opened 2017-02-28
Remove all artifacts on talk-deletion

heiglandreas:
When a talk is deleted now all associated artifacts are deleted also:

* Links to tracks
* Speakers
* Associated Links


heiglandreas:
This addresses (and resolves) https://github.com/joindin/joindin-web2/pull/400#issuecomment-281160665

PR #408, opened 2017-03-05
Trusted users

liam-wiltshire:
This is a solution for a couple of problems:

- Users who arrange long-running usergroups but who have to wait for approval every month
- Conference organising 'veterans' who have been running conferences for years, and for who we don't really need to spend time reviewing

A user can be set as 'trusted' by a joind.in admin, at which point their future event submissions get approved automatically (in the same way that admin submissions do).

There isn't really a set criteria for allowing someone to be trusted - I know there are a few conference/UG organisers who have been after this feature for a while and that we know well, so I guess we'd try it out on them first.

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -408,7 +408,7 @@ class UsersController extends ApiController
         $view = $request->getView();
         $view->setHeader('Content-Length', 0);
         $view->setResponseCode(204);
-        return;
+        return null;
     }
 
     public function setUserMapper(UserMapper $user_mapper)
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -402,7 +402,7 @@ class UsersController extends ApiController
 
         $user_mapper = $this->getUserMapper($db, $request);
         if (!$user_mapper->isSiteAdmin($request->getUserId())) {
-            throw new Exception("You must be an admin to change a user's trusted state", 401);
+            throw new Exception("You must be an admin to change a user's trusted state", 403);
         }
 
         $userId = $this->getItemId($request);
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -389,13 +389,10 @@ class UsersController extends ApiController
     /**
      * Allow users to be set as trusted
      *
-     * @param $request
-     * @param $db
-     *
-     * @return null
+     * @param $request Request
+     * @param $db      PDO
      *
      * @throws Exception
-     *
      */
     public function setTrusted($request, $db)
     {
@@ -414,12 +411,11 @@ class UsersController extends ApiController
         }
 
         if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
-            throw new Exception("Unable to update status");
+            throw new Exception("Unable to update status", 500);
         }
         $view = $request->getView();
         $view->setHeader('Content-Length', 0);
         $view->setResponseCode(204);
-        return null;
     }
 
     public function setUserMapper(UserMapper $user_mapper)
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -386,6 +386,31 @@ class UsersController extends ApiController
         $view->setResponseCode(204);
     }
 
+    public function setTrusted($request, $db)
+    {
+        if (false == ($request->getUserId())) {
+            throw new Exception("You must be logged in to change a user account", 401);
+        }
+
+        $user_mapper = $this->getUserMapper($db, $request);
+        if (!$user_mapper->isSiteAdmin($request->getUserId())) {
+            throw new Exception("You must be an admin to change a user's trusted state", 401);
+        }
+
+        $userId = $this->getItemId($request);
+        if (!is_bool($trustedStatus = $request->getParameter("trusted", null))) {
+            throw new Exception("You must provide a trusted state", 400);
+        }
+
+        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)){
+            throw new Exception("Unable to update status");
+        }
+        $view = $request->getView();
+        $view->setHeader('Content-Length', 0);
+        $view->setResponseCode(204);
+        return;
+    }
+
     public function setUserMapper(UserMapper $user_mapper)
     {
         $this->user_mapper = $user_mapper;
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -386,6 +386,17 @@ class UsersController extends ApiController
         $view->setResponseCode(204);
     }
 
+    /**
+     * Allow users to be set as trusted
+     *
+     * @param $request
+     * @param $db
+     *
+     * @return null
+     *
+     * @throws Exception
+     *
+     */
     public function setTrusted($request, $db)
     {
         if (false == ($request->getUserId())) {
@@ -402,7 +413,7 @@ class UsersController extends ApiController
             throw new Exception("You must provide a trusted state", 400);
         }
 
-        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)){
+        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
             throw new Exception("Unable to update status");
         }
         $view = $request->getView();
--- a/src/models/UserMapper.php
+++ b/src/models/UserMapper.php
@@ -38,6 +38,7 @@ class UserMapper extends ApiMapper
             "username"         => "username",
             "full_name"        => "full_name",
             "twitter_username" => "twitter_username",
+            "trusted"          => "trusted"
         );
 
         return $fields;
@@ -99,7 +100,7 @@ class UserMapper extends ApiMapper
     protected function getUsers($resultsperpage, $start, $where = null, $order = null)
     {
         $sql = 'select user.username, user.ID, user.email, '
-               . 'user.full_name, user.twitter_username, user.admin '
+               . 'user.full_name, user.twitter_username, user.admin, user.trusted '
                . 'from user '
                . 'left join user_attend ua on (ua.uid = user.ID) '
                . 'where active = 1 ';
@@ -226,6 +227,39 @@ class UserMapper extends ApiMapper
 
         return false;
     }
+
+    /**
+     * Check if the user represented by $user_id is trusted
+     *
+     * @param $user_id
+     * @return bool
+     */
+    public function isTrusted($user_id)
+    {
+        $results = $this->getUsers(1, 0, 'user.ID=' . (int) $user_id, null);
+        if (isset($results[0]) && $results[0]['trusted'] == 1) {
+            return true;
+        }
+
+        return false;
+    }
+
+    public function setTrustedStatus($trustedStatus, $user_id)
+    {
+        $verify_sql = "update user set trusted = :trusted_status "
+            . "where ID = :user_id";
+
+        $verify_stmt = $this->_db->prepare($verify_sql);
+        $verify_data = array("trusted_status" => (int) $trustedStatus, "user_id" => $user_id);
+
+        $result = $verify_stmt->execute($verify_data);
+
+        if ($result) {
+            return true;
+        }
+        return false;
+    }
+
 
     public function createUser($user)
     {
--- a/src/models/UserMapper.php
+++ b/src/models/UserMapper.php
@@ -244,6 +244,14 @@ class UserMapper extends ApiMapper
         return false;
     }
 
+    /**
+     * Update the trusted status for the given user
+     *
+     * @param $trustedStatus bool
+     * @param $user_id       int
+     *
+     * @return bool
+     */
     public function setTrustedStatus($trustedStatus, $user_id)
     {
         $verify_sql = "update user set trusted = :trusted_status "
@@ -252,12 +260,7 @@ class UserMapper extends ApiMapper
         $verify_stmt = $this->_db->prepare($verify_sql);
         $verify_data = array("trusted_status" => (int) $trustedStatus, "user_id" => $user_id);
 
-        $result = $verify_stmt->execute($verify_data);
-
-        if ($result) {
-            return true;
-        }
-        return false;
+        return $verify_stmt->execute($verify_data);
     }
 
 
--- a/tests/controllers/UsersControllerTest.php
+++ b/tests/controllers/UsersControllerTest.php
@@ -226,4 +226,192 @@ class UsersControllerTest extends \PHPUnit_Framework_TestCase
 
         $controller->updateUser($request, $db);
     }
+
+    /**
+     * Ensures that if the setTrusted method is called and no user_id is set,
+     * an exception is thrown
+     *
+     * @return void
+     *
+     * @test
+     * @expectedException        \Exception
+     * @expectedExceptionMessage You must be logged in to change a user account
+     */
+    public function testSetTrustedWithNoUserIdThrowsException()
+    {
+        $request = new \Request([], ['REQUEST_URI' => "http://api.dev.joind.in/v2.1/users/4/trusted", 'REQUEST_METHOD' => 'POST']);
+
+        $usersController = new \UsersController();
+        $db = $this->getMockBuilder('\JoindinTest\Inc\mockPDO')->getMock();
+
+        $usersController->setTrusted($request, $db);
+    }
+
+
+    /**
+     * Ensures that if the setTrsuted method is called and user_id is a,
+     * non-admin, an exception is thrown
+     *
+     * @return void
+     *
+     * @test
+     * @expectedException        \Exception
+     * @expectedExceptionMessage You must be an admin to change a user's trusted state
+     */
+    public function testSetTrustedWithNonAdminIdThrowsException()
+    {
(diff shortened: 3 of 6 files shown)


src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+    /**
+     * Allow users to be set as trusted
+     *
+     * @param $request
+     * @param $db
+     *
+     * @return null
+     *
+     * @throws Exception
+     *
```
heiglandreas: This line could be removed

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
         $view->setResponseCode(204);
     }
 
+    /**
+     * Allow users to be set as trusted
+     *
+     * @param $request
```
heiglandreas: $request is of type ```Request```. Shouldn't that be added here as type?

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+     * @param $db
```
heiglandreas: $db is of type PDO. Shouldn't that be added here as type?

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+     *
+     * @return null
```
heiglandreas: A return isn't needed. Should that not be ```void```?

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+            throw new Exception("You must provide a trusted state", 400);
+        }
+
+        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
+            throw new Exception("Unable to update status");
+        }
+        $view = $request->getView();
+        $view->setHeader('Content-Length', 0);
+        $view->setResponseCode(204);
+        return null;
```
heiglandreas: This isn't needed.

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+     */
+    public function setTrusted($request, $db)
+    {
+        if (false == ($request->getUserId())) {
+            throw new Exception("You must be logged in to change a user account", 401);
+        }
+
+        $user_mapper = $this->getUserMapper($db, $request);
+        if (!$user_mapper->isSiteAdmin($request->getUserId())) {
+            throw new Exception("You must be an admin to change a user's trusted state", 401);
```
heiglandreas: Should that not be a 403 - Unaouthorized?

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+        }
+
+        $userId = $this->getItemId($request);
+        if (!is_bool($trustedStatus = $request->getParameter("trusted", null))) {
```
heiglandreas: What are you checking here? ```is_bool``` checks whether the parameter is either ```true``` or ```false```. But ```$request->getPArameter()``` always returns a string… So ```is_bool``` should always fail here… Or am I missing something? 

I'd add a check whether 'trusted' is either "true" or "false" (or "yes" or "no")

src/controllers/UsersController.php
```diff
@@ -386,6 +386,42 @@ public function deleteUser($request, $db)
...
+            throw new Exception("You must provide a trusted state", 400);
+        }
+
+        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
+            throw new Exception("Unable to update status");
```
heiglandreas: I'd also add a 400-state here

src/models/UserMapper.php
```diff
@@ -227,6 +228,39 @@ public function isSiteAdmin($user_id)
...
+    {
+        $results = $this->getUsers(1, 0, 'user.ID=' . (int) $user_id, null);
+        if (isset($results[0]) && $results[0]['trusted'] == 1) {
+            return true;
+        }
+
+        return false;
+    }
+
+    public function setTrustedStatus($trustedStatus, $user_id)
```
heiglandreas: Can we have a DocBlock here as well?

src/models/UserMapper.php
```diff
@@ -227,6 +228,39 @@ public function isSiteAdmin($user_id)
...
+    {
+        $verify_sql = "update user set trusted = :trusted_status "
+            . "where ID = :user_id";
+
+        $verify_stmt = $this->_db->prepare($verify_sql);
+        $verify_data = array("trusted_status" => (int) $trustedStatus, "user_id" => $user_id);
+
+        $result = $verify_stmt->execute($verify_data);
+
+        if ($result) {
```
heiglandreas: As [PDO_Statement::execute()](http://php.net/manual/de/pdostatement.execute.php) always returns a boolean, wouldnt ```return $verify_stmt->execute($verify_data);``` be enough?

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+     *
+     * @return void
+     */
+    public function testSetTrustedWithSuccessCreatesView()
+    {
+        $request = $this->getMockBuilder('\Request')->disableOriginalConstructor()->getMock();
+        $request->method('getUserId')->willReturn(2);
+        $request->method('getParameter')
+            ->with("trusted")
+            ->willReturn(true);
```
heiglandreas: get parameter will return a string. So this should be ```->willReturn('0')```

src/models/UserMapper.php
```diff
@@ -38,6 +38,7 @@ public function getVerboseFields()
             "username"         => "username",
             "full_name"        => "full_name",
             "twitter_username" => "twitter_username",
+            "trusted"          => "trusted"
```
heiglandreas: Does this need to be exposed? We don't expose whether someone is an admin, so why exposing this? Or am I missing something?

src/controllers/UsersController.php
liam-wiltshire: Where do you get that? `$request->getParameter()` just returns something from the $pequest->parameters property, which in turn is the json_decode() of the Raw body... if the JSON includes a bool true or false (which is how I've designed the endpoint to work), then the return value of getParameter() will also be a bool :)

src/controllers/UsersController.php
liam-wiltshire: 403 is forbidden, 401 is Unauthorized - I've never really seen a consensus opinion one way or the other - as far as I understand it, the RFC implies that if a 401 is sent, then correct authentication may help - which means in both these situations a 401 is correct,

From the RFC:

 > If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials.

I don't have an issue with changing them, I just re-used an authentication snippet from further up the controller ;-)

src/controllers/UsersController.php
liam-wiltshire: I should be returning a code, however it should probably be in the 5xx range - this error will happen if the DB has refused to update the field for some reason, so that's going to be a server error rather than a user error - I'm thinking a nice, vague 500 :p

src/models/UserMapper.php
liam-wiltshire: There was a reason I exposed it - I think so that we (or indeed a user) can check if they are trusted or not. Otherwise the only way to know would be to dig in the DB, which obviously isn't ideal.

src/controllers/UsersController.php
heiglandreas: https://github.com/joindin/joindin-api/blob/master/src/inc/Request.php#L191states that the return-value is a string ;)

Looks like I missed the json_decode-part ;) Sorry…

src/controllers/UsersController.php
heiglandreas: As the user already *is* authorized (otherwise she would have been thrown out a few lines above) the action of adding the trusted-flag is not possible for the currently logged in user. I see 401 as "authentication-issue" whereas 403 is an "authorization-issue" (like http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses)

But I don't mind either. I'd connect this to https://joindin.jira.com/browse/JOINDIN-709 and https://joindin.jira.com/browse/JOINDIN-582 and leave it as it is ;)

src/controllers/UsersController.php
heiglandreas: Sounds like a good plan 😄

src/models/UserMapper.php
heiglandreas: I'm just thinking of handling being admin and being trusted the same way. 

Currently the "trusted" state can be seen by *any* user GETing the users information. But the information of being an admin is not available.

On the other hand, exposing that information only when the user itself is retrieving it is something completely different…

liam-wiltshire: I think it's more than just the user itself - any admin might need to be able to check the trusted state potentially.

I guess a nicer way to do it would be to only have the trusted field if the user is a site admin or the user, however AFAIK we don't have a clean way of defining mapper fields dependent on role? As such, I don't think exposing "trusted" is an issue from a security POV, so perhaps the bigger case is that we need to have a way of defining "admin" fields which are only returned whan the authenticated user has admin access to the current object?

liam-wiltshire:
I think I've responded and/or resolved everything now - not my comment on the "trusted" field in the mapper - if you agree, then suggest this needs to go into JIRA.

As such, if you are happy with the fixes @heiglandreas please confirm :-)

src/controllers/UsersController.php
liam-wiltshire: Added to JIRA as a quickfix :-)

liam-wiltshire:
Over to you @railto  :-)

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
         $controller->updateUser($request, $db);
     }
+
+    /**
+     * Ensures that if the setTrusted method is called and no user_id is set,
+     * an exception is thrown
+     *
+     * @return void
+     *
+     * @test
```
exussum12: Tiny but this line does nothing. With or without it the test is run due to the prefix.  Can it be one of the other (prefix or annotation)

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+    }
+
+
+    /**
+     * Ensures that if the setTrsuted method is called and user_id is a,
+     * non-admin, an exception is thrown
+     *
+     * @return void
+     *
+     * @test
```
exussum12: Can the error code be checked here. Agueably more important than the message

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+            ->with(204)
+            ->willReturn(true);
+
+        $request->expects($this->once())
+            ->method("getView")
+            ->willReturn($view);
+
+        $usersController = new \UsersController();
+        // Please see below for explanation of why we're mocking a "mock" PDO
+        // class
```
exussum12: Maybe I've missed this due to looking on a phone but it can't see where the explanation for this is

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+
+        $userMapper
+            ->expects($this->once())
+            ->method('isSiteAdmin')
+            ->will($this->returnValue(true));
+
+        $userMapper
+            ->expects($this->once())
+            ->method("setTrustedStatus")
+            ->willReturn(true);
```
exussum12: For consistency the above call should be the same (shorthand version)

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+
+        $usersController = new \UsersController();
+        // Please see below for explanation of why we're mocking a "mock" PDO
+        // class
+        $db = $this->getMockBuilder('\JoindinTest\Inc\mockPDO')->getMock();
+
+        $userMapper = $this->getMockBuilder('\UserMapper')
+            ->setConstructorArgs(array($db,$request))
+            ->getMock();
+
```
exussum12: For readability here i would make a new method which just sets the admin. The test becomes much more readable IMO

src/controllers/UsersController.php
```diff
@@ -386,6 +386,38 @@ public function deleteUser($request, $db)
         $view->setResponseCode(204);
     }
 
+    /**
+     * Allow users to be set as trusted
+     *
+     * @param $request Request
```
exussum12: These are type hinted here but not in the function itself. Can they be added there too please

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+     * @return void
+     *
+     * @test
+     * @expectedException        \Exception
+     * @expectedExceptionMessage You must provide a trusted state
+     */
+    public function testSetTrustedWithoutStateThrowsException()
+    {
+        $request = $this->getMockBuilder('\Request')->disableOriginalConstructor()->getMock();
+        $request->method('getUserId')->willReturn(2);
```
exussum12: Does this need to be mocked ? All of the other calls are not mocked ?

src/controllers/UsersController.php
heiglandreas: The ```$request``` and ```$pdo``` are type-hinted in a very inconsistent way throughout the method-calls. So leaving them without a type-hint would be OK from my POV but we then should perhaps add an issue to JIRA to go through all the Controllers and change that.

tests/controllers/UsersControllerTest.php
```diff
@@ -226,4 +226,192 @@ public function testThatUserDataIsNotDoubleEscapedOnUserEdit()
...
+            ->willReturn(true);
+
+        $request->expects($this->once())
+            ->method("getView")
+            ->willReturn($view);
+
+        $usersController = new \UsersController();
+        // Please see below for explanation of why we're mocking a "mock" PDO
+        // class
+        $db = $this->getMockBuilder('\JoindinTest\Inc\mockPDO')->getMock();
```
heiglandreas: @liam-wiltshire Why exactly are we using the ```\JoindinTest\Inc\mockPDO```-class in tests? As 
```$db = $this->getMockBuilder('\PDO')->disableDefaultConstructor()->getMock()``` does the same without needing a custom class…

exussum12: Presumably because of http://stackoverflow.com/questions/3138946/mocking-the-pdo-object-using-phpunit though i agree we should have that and a set request and DB up in the set up. As it's used everywhere.

heiglandreas: That question targets PHPUnit 3.x, joindin uses PHPUnit 5.x - and we haven't had issues with it so far ([Event_hostControllerTest](https://github.com/joindin/joindin-api/blob/master/tests/controllers/Event_hostsControllerTest.php) or [EventHostMapperTest](https://github.com/joindin/joindin-api/blob/master/tests/models/EventHostMapperTest.php))

exussum12: Yeah but when answers are old they tend to become the way people do things rather than the new ways :( 

That probably is new ticket worthy as the chnage to remove will be fairly big

tests/controllers/UsersControllerTest.php
liam-wiltshire: This is mocked to test/handle the getParameter call - in the other tests we never get to this point, so it's not needed

tests/controllers/UsersControllerTest.php
liam-wiltshire: Personally I normally use the prefix over the annotation, but most of the already existing tests there is both, so I've gone with both for consistency

liam-wiltshire:
Have pushed a few changes to clean the code up a little and respond to some of the concerns, so @railto   I guess it's over to you if everyone is happy :)

src/models/UserMapper.php
heiglandreas: Sounds like a good idea. I created Issue https://joindin.jira.com/browse/JOINDIN-734 to follow up on that

src/controllers/UsersController.php
heiglandreas: Added https://joindin.jira.com/browse/JOINDIN-735 to be able to follow up on that

railto:
Tried to run this locally, but unfortunately getting - https://www.dropbox.com/s/3cxukd4mal4x4mg/Screenshot%202017-03-27%2015.22.00.png?dl=0

PHPUnit also throwing some 70 errors when run via phing on the vagrant box

railto:
Never mind, local issue of my own doing

PR #409, opened 2017-04-23
Adds a first draft of a calendaring-interface

heiglandreas:
This allows to use https://api.joind.in/v2.1/events as source for a
calendaring application to display the events available on joind.in


heiglandreas:
Failing tests are due to SabreDav (which is required for creating iCalendar-Files) not being added to the vendor-folder.

Shall we add a composer-enabled workflow? Or still add the libs to the vendor folder?

railto:
I Know it had been discussed before and people opted to keep the current workflow of committing the vendor dir, however I personally would like to see this end.

PR #410, opened 2017-04-25
use composer for deployment

heiglandreas:
This PR adds ```composer install``` to the deployment process for the API.

This should be fairly safe to use as only libraries that are already defined in the composer.lock-file are added to the website and should a library be unavailable during the launch-process the script will break there and the incomplete build will not go live.

This PR does **not** remove libs that are already within the vendor-folder but it will overwrite the autoloading-informations.

This is though as a base for discussion.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+.gitattributes export-ignore
+.gitignore export-ignore
+scripts export-ignore
+tools export-ignore
+vendor/* export-ignore
+build.xml export-ignore
+build export-ignore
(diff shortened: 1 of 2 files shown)


railto:
As discussed previously on IRC, I agree with using composer as part of our deploy process and I think this is a perfect way to ease into it, maybe once everything is working fine with composer in the deploy then we can look at removing /vendor from git.

Can we please start a discussion on this in this PR so that we can have a log on it.

liam-wiltshire:
So, I tried this to see what would happen :-) - we don't have composer installed on the production server, so it broke :-).

So, I guess we need to update something to include composer - either include it on the server, or pull it in as part of the build step.

Thoughts?

PR #412, opened 2017-05-26
Replace the speaker_name when a talk claim is approved

liam-wiltshire:
Fixes https://joindin.jira.com/browse/JOINDIN-745

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -573,6 +573,7 @@ class TalksController extends BaseTalkController
         if (! $speaker_id) {
             throw new Exception("Specified user not found", 404);
         }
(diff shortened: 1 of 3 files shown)


heiglandreas:
While this fixes https://joindin.jira.com/browse/JOINDIN-745 it does not fix https://joindin.jira.com/browse/JOINDIN-742 though the reasons are similar…

Wouldn't a more generic approach be more interesting? Like setting the user-name in the edit-field of approved talks from the speakers name and not from the name field?

liam-wiltshire:
Yes, I agree in principle, but we needed a quick fix as speakers were losing claimed talks - affected 5 speakers at tek for example.

heiglandreas:
No Problem at all! I'm already working on a more general fix…

heiglandreas:
Find a generic fix in PR #413

PR #413, opened 2017-05-27
Fixes changed users name on editing talks

heiglandreas:
As soon as a user changed the name and then edited a talk they where already assigned to the new name would appear in the speaker-name field but on submitting the already claimed talk was reassigned to the user with the new name. That happened due to the API comparing the users name
to the value stored in the talk_claim table that might not be valid anymore. Therefore that value no longer is checked on editing a talk.

This solves JOINDIN-743 and JOINDIN-745

Tests still to come!

### DIFF
--- a/src/models/TalkMapper.php
+++ b/src/models/TalkMapper.php
@@ -490,11 +490,10 @@ class TalkMapper extends ApiMapper
     public function updateSpeakersOnTalk($talk_id, array $speakers)
     {
         // get the current speakers
-        $sql = 'select ts.speaker_name, ts.speaker_name as val from talk_speaker ts where ts.talk_id = :talk_id';
+        $sql = 'select if(ts.speaker_id,user.full_name, ts.speaker_name) as val from talk_speaker ts left join user on user.ID = ts.speaker_id where ts.talk_id = :talk_id';
         $stmt = $this->_db->prepare($sql);
         $stmt->execute(['talk_id' => $talk_id]);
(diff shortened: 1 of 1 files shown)


liam-wiltshire:
This doesn't fix the problem with changing your own name however? Just tested on dev:

1. Claimed Talk
2. Approved Claim
3. Edited Talk
4. Changed Name from Ima Admin => Ima J Adminson
5. Talk is no longer assigned, and Ima J Adminson shows as an unclaimed speaker

I think, fundamentally we need to change how name editing works:
 - Either claimed speakers should not be able to edit the display name (but this needs to follow through from the UI to the API)
 - The speaker can change the display name, but the speaker claim should still be in place...

heiglandreas:
The change addresses the following issue: 
* User changes her name via user/[username]/edit
* Much later the user edits a talk she is an approved speaker on to add slides
* The "speakers" field now contains the new name
* After submitting the slides are added but the user is no longer an approved speaker

From what I unserdtood you tested the other way around: The user edits the speaker name (along with adding the slides-link) **on the talk** - That of course should cause the current user to be unassigned and a new user to be added to the talk. That's in my eyes expected behaviour.

The issue addressed with this PR is that the preset speaker name of the talks edit-form previously contained the name from the "talk_speakers"-table (which is not changed when a user edits their name via /user/[username]/edit) and now contains the users name as set in the "users"-table

PR #414, opened 2017-06-04
Fixes Emojis not being storable

heiglandreas:
This change was made to allow storage of emojis as most relevant database fields where using the utf8 encoding and not the utf8mb4

This will be changed by this commit

### DIFF
--- a/src/public/index.php
+++ b/src/public/index.php
@@ -46,11 +46,6 @@ $ji_db = new PDO(
 );
 $ji_db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
-// Set the correct charset for this connection
-$ji_db->query("SET NAMES 'utf8' COLLATE 'utf8_general_ci'");
-$ji_db->query("SET CHARACTER SET 'utf8mb4'");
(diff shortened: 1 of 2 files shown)


src/public/index.php
```diff
@@ -40,16 +40,16 @@ function handle_exception($e)
...
     'mysql:host=' . $db['default']['hostname'] .
-    ';dbname=' . $db['default']['database'],
+    ';dbname=' . $db['default']['database'] . ';charset=utf8mb4',
     $db['default']['username'],
     $db['default']['password']
 );
 $ji_db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
 // Set the correct charset for this connection
 $ji_db->query("SET NAMES 'utf8' COLLATE 'utf8_general_ci'");
```
bcremer: All three of those queries should be unnecessary. Setting the charset during PDO connection time in line 43 is sufficient.

heiglandreas: That is correct! I was so busy fixing the bad that "SET NAMES utf8" did that I didn't realize I could remove that as well ;)

Thanks!

PR #417, opened 2017-06-14
Allow users to create a login via twitter

heiglandreas:
This change allows users to not only log in via twitter but also to create a new account via login in from twitter.

This should make it much easier to get users to log into joind.in as there's much less effort to be taken to get an account.

That should raise the adopition rate for conference-attendees.

When a user now tries to log in via twitter and there is no such user-account a new account is created and the user is logged in. As the Email-address is taken from twitter and they already validated it, we (IMO) do not need a second validation step so we can let the user do whatever they want right away. Perhaps that might need to be tweaked for users that do not have set an email-address in twitter…

This targets https://joindin.jira.com/browse/JOINDIN-741

### DIFF
--- a/src/controllers/TwitterController.php
+++ b/src/controllers/TwitterController.php
@@ -116,7 +116,7 @@ class TwitterController extends ApiController
                 try {
                     $res = $client1->get('1.1/account/verify_credentials.json?include_email=true');
 
-                } catch(Exception $e) {
+                } catch (Exception $e) {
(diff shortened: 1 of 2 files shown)


railto:
Looks good.  Don't have time to do a full review on it until later, if no one has picked it up by then i'll take a proper look

liam-wiltshire:
Working for me :-)

PR #418, opened 2017-06-29
Cannot TRUNCATE talks in db seeding

giorgiosironi:
`db/patch70.sql` introduces foreign keys that point to talks. When there are foreign keys present, TRUNCATE fails because it is a command that cannot check the data for constraint violations. Therefore, use DELETE FROM and reset the auto increment value.


liam-wiltshire:
ok to test

PR #420, opened 2017-06-29
Stop filtering title

giorgiosironi:
Filtering strip away valid text like tags (`<blink>`). With https://github.com/joindin/joindin-web2/pull/416 front ends will escape correctly, so we can start allowing these titles.

### DIFF
--- a/src/controllers/TalksController.php
+++ b/src/controllers/TalksController.php
@@ -439,11 +439,7 @@ class TalksController extends BaseTalkController
         }
         $event = $list['events'][0];
 
-        $talk['title'] = filter_var(
-            $request->getParameter('talk_title'),
-            FILTER_SANITIZE_STRING,
-            FILTER_FLAG_NO_ENCODE_QUOTES
-        );
+        $talk['title'] = $request->getParameter('talk_title');
         if (empty($talk['title'])) {
             throw new Exception("The talk title field is required", 400);
         }


liam-wiltshire:
ok to test

heiglandreas:
I'm not really a fan of that! That means that the output of the API is no longer safe to consume for the client. That means we put the burden of escaping onto the client.

liam-wiltshire:
In https://joindin.jira.com/projects/JOINDIN/issues/JOINDIN-736?filter=allopenissues it was argued not to escape?

heiglandreas:
Exactly. We are not escaping the output. That means we need to sanitize the input. Otherwise malicious code (like javascript) can (and will) be inserted and will be output right away to the client. And then the client needs to make sure that that malicious code does not do harmful things…

liam-wiltshire:
But by sanitising the way we currently are, potentially valid talk titles will just have part of it randomly deleted :-) - *or* we escape things, or use entities (which is kind of the same thing)

heiglandreas:
Can we sanitize the input in a different way? Like removing potentially malicious code (script-tags, anchors with href?, stuff like that and leave everything else as it is?

heiglandreas:
But not sanitizing user-provided input at all somehow feels creepy…

giorgiosironi:
An alternative to filtering input (currently stripping away tags) would be to escape it, transforming `&lt;input&gt;` into `&lt;blink&gt;`. It doesn't seem like a good solution as it may break the edit form?

Another alternative would be to refuse the input that, when filtered to strip tags, gets modified. We would still not allow titles like `The magic of <blink>` but it would be reported to the user. It's a *validate* rather than *escape* input solution.

liam-wiltshire:
Closing this PR as I think the general consensus is that this isn't the correct solution. However it is something that still needs looking at.

There is a JIRA issue https://joindin.jira.com/browse/JOINDIN-739 - suggest we discuss potential solutions there rather than in a PR :)

PR #423, opened 2017-06-29
Solves the broken jenkins cleanly

heiglandreas:
This solves the broken jenkins by using the currently checked out scripts-directory instead of the scripts-directory in the web-directory. Having fixed that we can again remove the tools and the scripts
directory as well as the tests…

Please @liam-wiltshire Check this before deploying 😉

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,8 @@
 .gitattributes export-ignore
 .gitignore export-ignore
+scripts export-ignore
+tests export-ignore
+tools export-ignore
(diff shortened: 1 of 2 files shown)


liam-wiltshire:
This is where I wish I knew our jenkins process better - will merge and try to deploy on test, see how it goes :)... wish me luck

heiglandreas:
So wishing luck helped? ;)

liam-wiltshire:
Either that or the enchanted amulet I found :p

PR #426, opened 2017-08-09
Remove autogenerated methods based on the parent

ericpoe:
These methods will be run anyway, so why define them as just calling the parent?


liam-wiltshire:
ok to test

ericpoe:
Uh. What am I missing? The details of the failing check show nothing wrong.

heiglandreas:
Looks like PHPUnit throws up somehow…

liam-wiltshire:
Yeah, that's wierd .... taking a look :)

liam-wiltshire:
test this please

liam-wiltshire:
Ok, PHPUnit really doesn't like something - presumably it all tests fine on your local?

liam-wiltshire:
Unless I'm wrong (probably!),, I think it must be fatal-ing somewhere:

"Execution of target "phpunit" failed for the following reason: /var/lib/jenkins/jobs/api-pr-build/workspace/build.xml:21:38: Task exited with code 255"

Pretty sure 255 is a PHP Fatal Error... not sure why it's not throwing up the error itself though unless it's suppressed somehow - will pull the branch down and take a look later unless someone beats me to it!

ericpoe:
@liam-wiltshire Yes, it looks like phpunit is throwing an error. I don't know why I didn't see it before when running the test. Caching?

Anyway, `ApiMapperTester::testBuildingTheLimitClause` throws an error because `ApiMapper::buildLimit` is protected. Should it be public in order to test?

heiglandreas:
I knew there was a reason why I added those methods!! That's it. They are exposed as public while they still can be protected in the original class… :smile:

heiglandreas:
So we probably should close this PR?

ericpoe:
This should be commented. I'll open a new PR so that we can explain to future contributors (including future me) why these methods that call parent are needed.

heiglandreas:
That is a good idea as well! Thanks for that!

ericpoe:
See #431

PR #427, opened 2017-08-09
Fix exception message

ericpoe:
This fixes an exception error message so that the right global is referenced.


liam-wiltshire:
ok to test

PR #429, opened 2017-08-29
Explicitly declare $retval arrays

ericpoe:
In some locations, a `$retval` (aka "return value") array is implied but is never explicitly declared. Explicitly declaring the array has 2 benefits:
- it looks nicer
- it makes the IDE (in my case, PHPStorm) realize that what it is returning is actually an array and not some unknown type

This PR explicitly declares the array.


liam-wiltshire:
ok to test

PR #430, opened 2017-08-29
Reduce temp vars

ericpoe:
One-time variables in a return statement reduce readability and can take up system resources. This PR attempts to remove some temporary variables from the API

In some cases, the reduction of temporary variables has led to the reduction of nesting, which makes the code a little easier to follow. When reducing nesting, I have favored returning early.

I have attempted to make the commits atomic so that if a type of change is undesired, it can be excised from this PR.


liam-wiltshire:
ok to test

src/controllers/ApiController.php
```diff
@@ -26,15 +24,8 @@ public function getItemId($request)
...
-        if (isset($request->parameters['verbose'])
-            && $request->parameters['verbose'] == 'yes'
-        ) {
-            $verbose = true;
-        } else {
-            $verbose = false;
-        }
-
-        return $verbose;
+        return isset($request->parameters['verbose'])
```
heiglandreas: Wouldn't it be easier to read like this?

```php
if (! isset($request-parameters['verbose'])) {
    return false;
}

if ($request->parameters['verbose'] !== 'yes') {
    return false;
}

return true;
```

src/controllers/TalksController.php
```diff
@@ -647,10 +646,10 @@ public function setSpeakerForTalk(Request $request, PDO $db)
 
     private function getLinkUserDataFromRequest(Request $request)
     {
-        $talk = [];
-        $talk['display_name'] = trim($request->getParameter('display_name', ''));
-        $talk['username'] = trim($request->getParameter('username', ''));
-        return $talk;
+        return [
+            'display_name' => trim($request->getParameter('display_name', '')),
+            'username' => trim($request->getParameter('username', '')),
```
heiglandreas: Shouldn't the ```=>``` be aligned under each other?

src/inc/Request.php
```diff
@@ -192,12 +192,11 @@ public function setRouteParams(array $routeParams)
      */
     public function getParameter($param, $default = '')
     {
-        $value = $default;
         if (array_key_exists($param, $this->parameters)) {
-            $value = $this->parameters[$param];
+            return $this->parameters[$param];
```
heiglandreas: Which one is the "happy path"? I'd personally say when there's a parameter to be retrieved. So I'd put it the other way round and say 

```php
if (! array_key_exists($param, $this->parameters)) {
    return $default;
}

return $this->parameters[$param];
```

src/inc/Request.php
```diff
@@ -212,13 +211,12 @@ public function getParameter($param, $default = '')
     public function getUrlElement($index, $default = '')
     {
         $index   = (int) $index;
-        $element = $default;
 
         if (isset($this->url_elements[$index])) {
-            $element = $this->url_elements[$index];
+            return $this->url_elements[$index];
```
heiglandreas: Same here. The "Happy Path" should be the one where we can get a URL-Element. So I'd switch that like for ```getParameter```

src/inc/Request.php
```diff
@@ -653,11 +651,10 @@ public function getClientUserAgent()
      */
     public function getConfigValue($key, $default = '')
     {
-        $value = $default;
         if (array_key_exists($key, $this->config)) {
-            $value = $this->config[$key];
+            return $this->config[$key];
```
heiglandreas: Same here. The "Happy Path" should be the one where we can get a Config Value. So I'd switch that like for ```getParameter```

src/models/ApiMapper.php
```diff
@@ -64,13 +64,10 @@ protected function buildLimit($resultsperpage, $start)
...
             // special case, no limits
-            $limit = '';
-        } else {
-            $start = (int)$start;
-            $limit = ' LIMIT ' . $start . ',' . $resultsperpage;
+            return '';
         }
 
-        return $limit;
+        return ' LIMIT ' . intval($start) . ',' . $resultsperpage;
```
heiglandreas: Why not add the ```intval``` for ```$resultsperpage```?

And why using ```intval``` instead of ```(int)```?

src/controllers/ApiController.php
ericpoe: I find my way cleaner and easier to read, but I'm not one of the core maintainers of the project ;)

5925a24

src/controllers/TalksController.php
ericpoe: That would look good. 43bb031

src/inc/Request.php
ericpoe: That would be better. I just didn't want to make too many changes in this PR. :) 6411dd7

src/inc/Request.php
ericpoe: Updated.

src/inc/Request.php
ericpoe: Updated.

src/models/ApiMapper.php
ericpoe: Updated to use the int cast as in the original: 9c0ca7d. I had thought that there was extra evaluation within `intval`, but it appears to not be the case. https://3v4l.org/Fmlnk

PR #431, opened 2017-08-29
Explain why test methods are needed

ericpoe:
Two of these test methods get tested, the one that is removed does not. Essentially `TestApiMapper` serves as a wrapper for `ApiMapper` and might be better served as a mock or by using reflection.

This PR explains why 2 methods in this class are needed.


railto:
Ok to test

liam-wiltshire:
ok to test

heiglandreas:
Wouldn't it be more interesting to actually have a test for the method instead of removing the method?

ericpoe:
@heiglandreas I would be good with that. Shall I create an issue and close this ticket?

heiglandreas:
Sorry for the delay, but I think closing this and adding an issue is IMHO the better option…

railto:
I agree with @heiglandreas on this. I'm closing out this PR, please open an issue to discuss this further, thanks @ericpoe

PR #434, opened 2017-10-14
Fix facebook login

exussum12:
The old login didnt seem to work correctly, new behaviour is as follows:

If the person has an email address which matches their facebook, log in
to that account
if no email address is found create a new account with information given
    from the facebook oauth api


boscho87:
because i want to improve my review skills, i just added my opinion to this PR :). 

Read it or just ignore it 😄. And tell me if i should not review on this repo anymore

exussum12:
I can't see your comments but all would be welcome :)

liam-wiltshire:
ok to test

src/controllers/FacebookController.php
```diff
@@ -38,7 +40,7 @@ public function logUserIn($request, $db)
             'headers' => ['Accept' => 'application/json']
         ]);
 
-        $res = $client->get('https://graph.facebook.com/v2.3/oauth/access_token', [
+        $res = $client->get('https://graph.facebook.com/v2.10/oauth/access_token', [
```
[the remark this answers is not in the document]
exussum12: This is done with the next PR, So I have not done it here also to prevent a conflict

src/models/OAuthModel.php
```diff
@@ -425,28 +425,60 @@ public function createUserFromTwitterUsername($clientId, $values)
...
-            $accessToken = $this->newAccessToken($clientId, $userId);
+        $userId = $result['ID'];
+        if (!$userId) {
+            return false;
+        }
 
-            // we also want to send back the logged in user's uri
-            $userUri = $this->getUserUri($userId);
+        // create new token
+        $accessToken = $this->newAccessToken($clientId, $userId);
```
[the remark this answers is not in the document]
exussum12: Updated :)

src/models/OAuthModel.php
```diff
@@ -425,28 +423,59 @@ public function createUserFromTwitterUsername($clientId, $values)
...
+     * @param  string $userName         Username to be created if not found
      * @return array|false              Array of access token and user uri on success or false or failure
      */
-    public function createAccessTokenFromTrustedEmail($clientId, $email)
+    public function createAccessTokenFromTrustedEmail($clientId, $email, $fullName = '', $userName = '')
     {
-        $sql = "select ID from user "
-            . "where email = :email "
-            . "and verified = 1";
+        $sql = "
```
heiglandreas: Do we really need to split that over multiple lines? Otherwise we probably should also split the function-definition over multiple lines…

src/models/OAuthModel.php
```diff
@@ -425,28 +423,59 @@ public function createUserFromTwitterUsername($clientId, $values)
...
+        $stmt = $this->_db->prepare($sql);
+        $stmt->execute(
+            [
+                "email" => $email,
+                'fullName' => $fullName,
+                'userName' => $userName
+            ]
+        );
+
+        return [
```
heiglandreas: Don't we need a failback here? What happens when executing the statement fails? Perhaps because the user already exists?

src/models/OAuthModel.php
exussum12: It's much more readable as it currently is. 

I can add a test none of this class / area is currently tested and would need parts re writing to be testable though

src/models/OAuthModel.php
exussum12: The check for an existing user happens before this statement is run. It tries to use that user first. So probably no fall back needed as this would any fail if the SQL is invalid

PR #435, opened 2017-10-15
Enable GitHub logins

exussum12:
Also refactors the current facebook login to be more generic. Should be a fairly simple case of adding any more oath providers now

### DIFF
--- a/src/config.php.dist
+++ b/src/config.php.dist
@@ -31,6 +31,12 @@ $config =  array(
         'app_id' => '',
         'app_secret' => '',
     ),
+    'github' => array(
+        // set up at https://github.com/settings/developers
+        // configure url in OAuth Settings as /user/github-access on web2
+        'app_id' => '',
+        'app_secret' => '',
+    ),
     'features' => array(
         'allow_auto_verify_users' => false,
         'allow_auto_approve_events' => false,
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -414,6 +414,12 @@
         "verbs": ["POST"]
     },
     {
+        "path": "/github/token",
+        "controller": "GithubController",
+        "action": "logUserIn",
+        "verbs": ["POST"]
+    },
+    {
         "path" : "/applications$",
         "controller" : "ApplicationsController",
         "action" : "listApplications",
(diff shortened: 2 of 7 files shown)


exussum12:
The old way used email addresses, So I have kept it working like this, I would suggest basing it from the ID instead though. 

Thoughts ?

heiglandreas:
If we want to incorporate more social logins it might be a good idea to have a look at [SocialConnect](https://github.com/SocialConnect/auth). That might make it easier to use different connectors without having to worry about keeping the API up-to-date ourselfs. There‘s also HybridAuth, but from what I‘ve seen it‘s not that actively developed anymore...

heiglandreas:
Additionally using multiple logins might make it necessary to add a new DB-table to connect users and their different logins....

exussum12:
Yeah agreed. Which would use the I'd instead of the email. Wasn't sure if there was a reason email was used I just didn't know hence keeping it. 

Happy to move this to being Id based so we know where they came from.

exussum12:
100% agree. Update this pr then to make it I'd based and forget about the email ?

liam-wiltshire:
ok to test

heiglandreas:
We'd need a complete rewrite IMHO of the login-mechanism. That should contain 

* Move from one user/login per account to multiple users/logins per account
* Implement login using a 3rd party auth-provider. (So we don't need to keep track of the different implementations) - I'D recommend SocialAuth but any other is fine

Before we do that I'D not recommend adding more login-mechanisms. After we did that implementing more is a matter of adding a new config-part.

exussum12:
Agreed with the multiple logins per account. It's a fairly large change though and will effect all of the current ones. 

Currently (after changes in this pr)  it's only a small case of adding config and mapping the resulting array which is different. Not sure I see the benifits of using a third party in this case as oauth is a standard.

svpernova09:
Safe to close this PR?

exussum12:
Personally I think GitHub is a better login option than Facebook. This follows the existing way of doing things. But it would be better long term to replace the existing way with a library. 

Would need composer to be sorted to allow this

svpernova09:
Closing this out for now, I'd be happy to revisit adding oauth providers once we can refactor existing code out to a library instead of our own code.

PR #437, opened 2017-11-03
Fix legacy upload

exussum12:
Editing a talk in legacy doesnt show the correct link


exussum12:
Closing as there is a case this won't work in

I will add some tests in this area too

PR #438, opened 2017-11-05
Adds rating info from logged in user

heiglandreas:
This commit adds information about the rating of the currently logged in user for talks. That way we can show

* that the user already rated that talk and
* how the rating was in regard to the overall rating

That information will only be available when a user is logged in and only that users information will then be shown.

### DIFF
--- a/src/models/TalkMapper.php
+++ b/src/models/TalkMapper.php
@@ -42,6 +42,12 @@ class TalkMapper extends ApiMapper
                     $results[$key]['starred'] = false;
                 }
 
+                // Did the logged in user rate this talk?
+                $results[$key]['user_rating'] = false;
+                if (isset($this->_request->user_id)) {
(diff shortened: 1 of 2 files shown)


src/models/TalkMapper.php
```diff
@@ -42,6 +42,12 @@ public function processResults($results)
                     $results[$key]['starred'] = false;
                 }
 
+                // Did the logged in user rate this talk?
+                $results[$key]['user_rating'] = false;
+                if (isset($this->_request->user_id)) {
+                    $results[$key]['user_rating'] = $this->getUserRatingOnTalk($row['ID'], $this->_request->user_id);
```
boscho87: maybe you could stroe the `$row['id']` into a variable.

and maybe: 
```
$result['user_rating'] = $stuff;
$result['speakers'] = $otherstuff;
// etc.
$results[$key] = $result;
```
👍

railto:
Thanks @heiglandreas

PR #441, opened 2018-03-24
Fix for allowing big ints

exussum12:
Fixes web issue 433

### DIFF
--- a/src/controllers/FacebookController.php
+++ b/src/controllers/FacebookController.php
@@ -61,7 +61,7 @@ class FacebookController extends ApiController
                 ]
             ]);
             if ($res->getStatusCode() == 200) {
-                $data = json_decode((string)$res->getBody(), true);
+                $data = json_decode((string)$res->getBody(), true, 512, JSON_BIGINT_AS_STRING);
                 if (!array_key_exists('email', $data)) {
                     throw new Exception("Email address is unavailable", 403);
                 }
--- a/src/views/JsonView.php
+++ b/src/views/JsonView.php
@@ -47,7 +47,7 @@ class JsonView extends ApiView
 
     protected function scalarNumericCheck($key, $value)
     {
-        if (is_numeric($value) && ! in_array($key, $this->string_fields)) {
+        if (is_numeric($value) && ! in_array($key, $this->string_fields) && $value < PHP_INT_MAX) {
             return (float) $value;
         }
 


exussum12:
@liam-wiltshire this may need a clean up in the DB. Can you check the DB for any numbers in scientific notation ?

heiglandreas:
Related issue: https://joindin.jira.com/browse/JOINDIN-334?jql=text%20~%20%22facebook%22

exussum12:
It's not as easy on the app. Will need to change the Facebook set up too. It needs to point back to the API server. Currently it points to the web one. 

Don't mind doing the work to move it over for the app

heiglandreas:
Looks good from my POV
👍

lornajane:
If there's work going on with the Facebook integration (which I thought was broken for a whole) then look out for API changes. I got this from Facebook:

v2.5 upgrade reminder  

Your App, Joind.In, is currently accessing Graph API v2.5 which will reach the end of its 2-year lifetime on 12 April, 2018. We estimate 2 endpoints that Joind.In calls will be impacted by this change and may stop working after the automatic upgrade push. To ensure a smooth transition, please migrate all calls to Graph API v2.6 or higher. Use the API Upgrade Tool to understand exactly how this change will impact your app.


Note: If you do not see any impacted calls in the API Upgrade Tool, your app/code may not be affected by this upgrade.


Visit our changelog to see the full list of changes in all Graph API versions.


 You can view this and other Developer Notifications related to your app, in the App Dashboard.

exussum12:
Ok i will check that as well. Thanks

exussum12:
@lornajane does it give any more info ? 

https://github.com/joindin/joindin-api/blob/master/src/controllers/FacebookController.php#L49  uses 2.10

Web2 doesnt  have a version number does anything else use the Facebook API ?

lornajane:
If it's going to break anyway, removing it would be a very positive move in my opinion!

exussum12:
This doesnt need anything to change with the facebook app ? This is how we handle data back from facebook

railto:
Looking through the thread there seems to be some discussion on updating FB

exussum12:
Yeah but that's already been done here. The discussion was for legacy. Legacy can be updated but again doesn't need Facebook just needs to update code 

Everything else has been moved

railto:
Thanks @exussum12

exussum12:
This is only part of the fix. Web2 has the other part of the fix

PR #445, opened 2018-05-30
Adapt tests to work with PHPUnit > 5

heiglandreas:


exussum12:
Clearly not for this ticket but vendor should be removed at some point, Makes this look much bigger than it really is

heiglandreas:
I'm absolutely with you on that! Doing the deployment right with composer would make the whole codebase much smaller ;-)

One step at a time…

railto:
Had a chat with @heiglandreas and we've both agreed that this should be put on ice until #446 is taken care of

ericpoe:
Let's revisit after #446 is merged in.

heiglandreas:
Rebased onto the current master...

PR #446, opened 2018-05-30
Move deployment to composer

heiglandreas:
This PR removes the vendor directory completely and adapts the autoloading apropriately.


exussum12:
Quarter of a million lines deleted. Nice work!

railto:
My hesitation on this is that we'd also need to ensure the deploy scripts are updated.  Unfortunately I don't know anything about deployments to the test/production platform.  Would need input on this from @liam-wiltshire or someone else with deploy access

heiglandreas:
According to [this](https://github.com/joindin/joindin-api/blob/cf89b8999c81fa82547226501f83dc0445d55507/scripts/jenkinslaunch.sh#L43) we're already running composer on every deployment…

But I'd also prefer a double check with @liam-wiltshire ;-)

iansltx:
Whoops, I build something similar, with a few tweaks for PHP 7 compat. I'll see if I can apply those fixes here rather than doing a thing in parallel. For my own reference:

* swiftmailer to ^5.4.9
* php-markdown to ^1.8.0
* Fix Travis build

There are a few other changes we'll want to re-roll on top of this.

Jean85:
Why composer.lock has been removed? Shouldn't we lock the deps version that will be deployed?

heiglandreas:
I can't tell you anymore why I removed the composer.lock.... Perhaps because I used a different PHP-Version?

But you are right. It should definitely be included!

michaelcullum:
This should also handle moving `tests/diffFilter.phar` (#614) to a composer dep

Also, lets modify https://github.com/joindin/joindin-api/blob/master/.travis.yml#L9 to include 7.3 as this PR should (once it has the deps updates in) give us 7.3 compat.

heiglandreas:
I'd rather use PHARs instead of composer dependencies as that doesn't tie the dependencies of the project to dependencies of build-tools. But apart from that  that looks good.

Question is whether we also include 7.2 and 7.1 in the build-matrix...

svpernova09:
>I'd rather use PHARs instead of composer dependencies as that doesn't tie the dependencies of the project to dependencies of build-tools.

IMHO everything to build the app in the repo should be declared as a dependency so the app can be built from it's own repo. (Bundle everything needed to test, build, etc in composer)

heiglandreas:
That's why I'd love to be able to declare PHARs in the composer.json but well... :-)

Nevertheless: That ties the joind.in-dependencies to the dependencies of the build-tools. That does not need to be bad but it is an architectural choice that I'd not want to make...

iansltx:
I'm +1 on pulling the build deps in via composer, though I understand the downsides there. At the point where dev dependencies are causing issues with "real" deps we can re-evaluate.

michaelcullum:
I'm also +1 on bringing in build tool deps via composer.

> Question is whether we also include 7.2 and 7.1 in the build-matrix...
On joindin/joindin-web2 we just added 7.3 (alongside 5.6).

The intention I guess will be to run 7.3 on new infrastructure and not 7.1 or 7.2 and I don't see a real need to support more versions of PHP than necessary.

michaelcullum:
@iansltx I've done a few of those bits making it consistent with joindin/joindin-web2#750 which also accidentally included fixing travis errors.

Will leave the updating of swiftmailer/markdown etc. to you though.

heiglandreas:
So the remaining tasks here are updating swiftmailer and php-markdown? And then we can remove the WIP?

zghosts:
just one small remark for composer.json
```
"swiftmailer/swiftmailer": "^5.4",
"michelf/php-markdown": "^1.8",
```

heiglandreas:
Thanks for spotting @zghosts - Fixed

zghosts:
Now in the right spot....
Did you rerun composer update after the last change?
I'm getting:
"Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them."
composer update or composer update --lock should fix this

heiglandreas:
Now I did :grin:

Didn't see the reason for it before as the changes where not altering the version-strings but I didn't realize that (of course) the checksum changed and therefore the warning would appear.

Fixed now.

zghosts:
Apart from the WIP label I don't any issues with merging

iansltx:
@mcneely added the label, which no longer applies with the latest updates. Dropping the label now.

iansltx:
WIP label no longer applies at this point, as the reasons for being WIP have been resolved. Will review in a bit.

composer.json
```diff
@@ -1,9 +1,60 @@
 {
+    "name" : "joindin/joindin-api",
+    "description" : "The API-backend that powers joind.in",
+    "keywords" : ["joind.in", "events", "talks", "speaking", "feedback"],
+    "homepage" : "https://api.joind.in",
+    "license" : "",
```
Jean85:
```suggestion
    "license" : "BSD-3-Clause",
```

composer.json
```diff
@@ -1,9 +1,60 @@
...
+    "support" : {
+        "issues" : "https://joindin.jira.com",
```
Jean85:
```suggestion
        "issues" : "https://github.com/joindin/joindin-api/issues",
```

iansltx:
@Jean85 Please go ahead and commit the changes you suggested. GitHub is giving me a generic error when attempting to batch the two up and commit via the web UI.

heiglandreas:
I accepted the changes from @Jean85

svpernova09:
I have verified via local environment (joind.in-vm) that the changes here for Swiftmailer are good. Mail is being sent via info@joind.in as expected.

zghosts:
@mcneely I think we should make smtp authentication optional adding something like $config['email']['smtp']['authentication'] = false (default) and set security, username and password to null in config.php.dist
as the default smtp server is localhost on port 25, this by default does is not or has no need to be authenticated.

heiglandreas:
Does that need to be added in this already rather huge PR? Or is that something that can be added in a separate PR?

PR #449, opened 2018-07-03
Allow users to add a small biography to their profile

Pton:
This will allow a user to add some more information regarding their occupation, specialization, interests and experience. That additional information can be useful for users interested in a speaker, for event organizers to get a feel for a speaker or for getting some context behind a comment.

The changes made in the API will be used by joindin/joindin-web2#443.

Github pages documentation is updated in #450.

### DIFF
--- a/db/patch74.sql
+++ b/db/patch74.sql
@@ -1,0 +1,5 @@
+# adding the bio column to the user info
+
+ALTER TABLE `user` add `biography` VARCHAR(400);
+
+INSERT INTO patch_history SET patch_number = 74;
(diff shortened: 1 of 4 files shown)


db/patch74.sql
```diff
@@ -0,0 +1,5 @@
+# adding the bio column to the user info
+
+ALTER TABLE `user` add `biography` VARCHAR(400);
```
heiglandreas: Wouldn't a column type of `TEXT` be better here?. That way we don't add those 400*4 byte to the max row-length :wink:

exussum12: agree here. text would be nice.

Pton: Thanks for the review. I've changed the data type to `TEXT`

railto:
Thanks, this is a nice addition

PR #451, opened 2018-07-10
Add keyword-based search to users endpoint

Pton:
This will allow for improved search on the website. For instance a user will be able to look up talks for a speaker they notice on twitter. Additionally event organizers will be able to quickly navigate to the
user profile of a speaker in order to see previous their talks or attended events.


railto:
@Pton Thanks for submitting this.

PR #453, opened 2018-07-23
Show that users need to connect to legacy

BackEndTea:
joind.in does not have the user/oauth_allow route.
This route only exists in the legacy application

### DIFF
--- a/howto/authenticate.md
+++ b/howto/authenticate.md
@@ -21,7 +21,7 @@ This page is in two parts - the first section shows you which applications you h
 
 For each user, you will need to send them over to joind.in to log in as themselves and confirm that they grant your application access to their data.  The URL you want is:
 
-    http://joind.in/user/oauth_allow?api_key=[api-key]&callback=[callback_url]
+    http://legacy.joind.in/user/oauth_allow?api_key=[api-key]&callback=[callback_url]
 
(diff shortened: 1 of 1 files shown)


lornajane:
You are correct that these docs are out of date.  However the users should be using the endpoints on the API, _not_ the legacy site which will be turned off at some future time.  There should be docs on this somewhere but if not, the current website logs users in by this route so that could be a good example to look at.

akrabat:
The API docs are here: http://joindin.github.io/joindin-api/

BackEndTea:
> However the users should be using the endpoints on the API, not the legacy site which will be turned off at some future time.

The login through the current website ends up in this function if I'm correct.
https://github.com/joindin/joindin-api/blob/56baaac6e6969b15c9be60b0c8a2266485dc2271/src/controllers/TokenController.php#L9

That requires me to send along a username and password, instead of allowing the user to enter it there themselves.

lornajane:
These docs might be more help http://joindin.github.io/joindin-api/howto/authenticate.html - the website uses a different style of OAuth grant because it's a trusted client.  We might be more help if we knew what it was you wanted to do, feel free to share a bit more information about your project if you would like to?

lornajane:
Oh I love this idea!  We had something like this a while ago but it hasn't been active lately https://github.com/joindin/JoindIn.js

BackEndTea:
Of course, I just meant that the mentioned javascript library does not have that functionality.

akrabat:
Ah! Sorry. I misread.

PR #447, opened 2018-06-05
Hide the first ratings

heiglandreas:
Often it looks like the rating and the tone of the first comment influence the following ratings and comments. To allow people to rate and comment without being influenced too much this PR hides the comments until a certain threshold has been reached. Currently that is either more than 3 comments or the talk has been given more than 3 days ago. Until one of these thresholds has been reached there are no comments shown for the talk. 

Hopefully that will allow a more uninfluenced rating of talks…

PR #455, opened 2018-10-05
Add Travis checks

exussum12:
Currently only doing psr2 and phpunit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ before_script:
   - composer global require phpunit/phpunit
   - composer global require jakub-onderka/php-parallel-lint
   - composer install
-  - phpenv config-rm xdebug.ini 
   - cp src/config.php.dist src/config.php
 script:
 - PATH=$HOME/.composer/vendor/bin:$PATH ./build.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,6 @@ before_script:
   - composer global require jakub-onderka/php-parallel-lint
   - composer install
   - phpenv config-rm xdebug.ini 
+  - cp src/config.php.dist src/config.php
 script:
 - PATH=$HOME/.composer/vendor/bin:$PATH ./build.sh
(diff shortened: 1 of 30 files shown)


exussum12:
This will need someone to enable travis on the build. https://docs.travis-ci.com/user/getting-started/

First few steps of that

heiglandreas:
Currently we are using jenkins for the CI-pipeline at http://jenkins.joind.in

So far we did not (yet) decide to move away from that.

Though there are some ideas to change the projects setup there is no decission to actually change all that. We should schedule a chat session on IRC to discuss that stuff.

But as long as there is no decission to move CI from jenkins to github I would not invest time into such a move. 
B

exussum12:
Jenkins doesn't currently work (hence the reason for the psr2 pull request as it was already done) 

Some talk about it here 


https://github.com/joindin/joindin-api/pull/447#issuecomment-396944443

Personally I would prefer travis as it shows correctly on the commits

heiglandreas:
Personally I'm also :+1: on that. I would just prefer a team decission on the future strategy and vision of the whole project so that we can then actually move on strongly in one direction.

exussum12:
Happy be in involved in a chat. When's best to set one up ?

railto:
Moving away from Jenkins is part of a larger conversation around the infrastructure behind joindin. There are things being done in Jenkins that need to be looked into to see if they can be easily replicated in Travis or another service like CircleCI.

This is something that I'm personally going to be looking into over the coming weeks and once I have a solid grasp on what currently happens vs what needs to be done then I'd like to get a conversation going around this.

@exussum12 Thank's for submitting the PR.  I'm going to close it for now, but we may well end up using it in future, however, for now, I'd rather stick with Jenkins until we figure out the bigger picture

exussum12:
Nothing is done currently in Jenkins though due to it being broke for almost a year ?

What is here currently set up travis and then runs a shell script, that can be done easily for any provider we use. To me having something working is better than not having anything even if its just temporary ?

railto:
Per IRC discussion, re-opening this. Would love to see the same for Web2

PR #456, opened 2018-10-05
PSR2 check so we can add the build

exussum12:
This just runs

phpcbf --standard=tools/codesniffer/JoindInPSR2/ruleset.xml tests/

and

phpcbf --standard=tools/codesniffer/JoindInPSR2/ruleset.xml src


railto:
Thanks @exussum12

PR #458, opened 2018-10-07
Make the event comment match the talk comments

exussum12:
Talk comments at event level are descending. Event comments are ascending. This PR will make them both descending (most recent first)


exussum12:
Talks listed here
https://github.com/joindin/joindin-api/blob/master/src/models/TalkCommentMapper.php#L79

railto:
Thank you @exussum12

PR #608, opened 2018-10-25
Update patch75.sql

exussum12:


exussum12:
Does this allow the patch to run ?

railto:
yup, tested it and then ran, all good, thank you

PR #609, opened 2018-10-27
JOINDIN-763 #close

cal-tec:
Usernames were being presented in a dynamic type form. in the vent of a long numeric username being represented in a scientific number format when returned by the API. This was reported as an issue with web2 but the error actually existed within the API.

### DIFF
--- a/src/views/JsonView.php
+++ b/src/views/JsonView.php
@@ -22,7 +22,7 @@ class JsonView extends ApiView
         // Don't use JSON_NUMERIC_CHECK because it eats things (e.g. talk stubs)
 
         // Specify a list of fields to NOT convert to numbers
-        $this->string_fields = array("stub", "track_name", "comment");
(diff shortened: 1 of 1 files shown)


railto:
Thanks @cal-tec

exussum12:
@cal-tec unfortunately I think they were written to the DB incorrectly :(

cal-tec:
Following on from this issue I have made a number of PR's to remove some pointless (and most likely legacy) complexity from the code base. I am fairly sure [this PR](https://github.com/joindin/joindin-web2/pull/732/files) resolves the issues with caching that are mentioned above. Usernames were being cached by a key of the users URI but this is pointless as the API endpoint being called now provides the username on each request. The linked PR will, therefore, eliminate the cache in this instance and resolve the issue with the live version of web2 still not behaving as expected.

PR #440, opened 2018-02-22
Fix legacy slides

exussum12:

PR #610, opened 2018-10-27
Fixes issue with creating a talk.

railto:
Insert query had slides_link as field, but no value set for it.


exussum12:
Also I think I added tests around here. Any chance of a test adding

railto:
This fixes a bug that was introduced in https://github.com/joindin/joindin-api/pull/440/files, the value etc for slides_link was removed but the field not.

heiglandreas:
That probably would not have shown up in a unit-test. And the integration-test-suite is not really in a healthy state... :-/

exussum12:
Ah i see. I already handled the legacy slide media. I will experiment with dropping that column I don't think we will need it any more. 

Thanks for fixing @railto

PR #612, opened 2018-12-01
WIP: Add possibility to use ActionControllers

heiglandreas:
This PR will allow us to use not only the so far used controllers with multiple actions but it will also allow us to use Single-Action Controllers that are only responsible for one single thing. That will allow us to separate the concerns of a controller more easily and to also limit the actual controller-logik to just one thing.

While it means adding mor efiles to the codebase it will also mean adding less code to one file and therefore less possibilities to include errors. The single classes will also be easier to understand.

The whole feature can e used by defining a new route in `2.1.json` like this:

```json
{
    "path": "/url/path",
    "actioncontroller": "MyActionController",
    "verbs": ["GET"]
}
```


heiglandreas:
Ouch. This contains actually two different things and only mentions one in the PR-Title.

I'll close this for now as it was merely to begin the discussion whether ActionControllers would be a way to go for future development. I'll spark that discussion with a new PR that only contains the relevant code.

PR #614, opened 2019-02-02
Enforce test coverage

### LINKED ISSUE #577: JOINDIN-700: Addition of Unit Tests on API
Jira issue originally created by user liam-wiltshire:

At the moment, there are not many Unit Test, particularly we could do with adding some on the API side.

Because of how some if it is structured (calling header() / exit() mostly), there may be some code changes required to allow tests to be written.

exussum12:
Require each PR to have at least 80% unit test coverage

fixes #577


exussum12:
Added as a phar due to the composer issue not being sorted, when that PR is moved this can be moved to composer instead. Trying to avoid a merge issue

PR #616, opened 2019-02-03
Upgrade SwiftMailer, Markdown libs, drop /vendor from version control

iansltx:
This gets us PHP 7.x support and lib versions that don't have CVEs for the ones I bumped. This is as far as we can go on SwiftMailer without requiring PHP 7, so bumping to SwiftMailer 6.x will be its own PR.

PHPCompatibility hands back what appears to be a false positive on Guzzle's event emitter (tested the flagged behavior in 7.2 and had no issues), and flags various magic_quotes refs in SwiftMailer that are behind function_exists checks, as well as an ini_get for safe_mode (which returns false in 7.x...no fatals or anything). None of these issues will prevent us from moving all the way to 7.3 here (prod's on 5.6 now).


iansltx:
Never mind. Should just build on https://github.com/joindin/joindin-api/pull/446 instead.

PR #620, opened 2019-02-06
Add POST /users/:id/admin endpoint (works similar to /trusted)

iansltx:
**NOTE:** Untested-in-real-life. Tests are written for the controller and should cover the same % of code that setTrusted does.


heiglandreas:
Obsoleted by #675

Issue #621, opened 2019-02-06
[Discussion] User self-deletion procedures


iansltx:
Right now, to delete themselves users have to either request deletion via a contact form that doesn't authenticate them, or else hit an API endpoint that we'll be stripping non-admin permissions from.

Copying from chat discussion on this by @zghosts: The simplest solution for now is to add "request deletion" and dissallow self-deletion both through the api and the ui, that would cover the legal requirements for GDPR in terms of "the right to be forgotten".

A further expansion would be to self-delete with a quarantine period, representing as an anonymous user in the ui, before being permanently deleted and anonimised after a week or two or manually by an admin.

Question is what which steps we want to take here, what exactly that looks like, and when those steps get taken, given that the contact form method, though manual, is technically sufficient, particularly given that admin-based user deletion fully works.

svpernova09:
IMHO
Short term solution: Disable self deletion via API
Long(er) term solution: pair UI & API functionality to allow users to delete themselves according to whatever guidelines we decide in this thread.

iansltx:
Okay; closing https://github.com/joindin/joindin-web2/pull/756 as a result of this.

iansltx:
See also: https://github.com/joindin/joindin.github.io/blob/master/GDPR.md

PR #622, opened 2019-02-06
Disallow users from deleting themselves

iansltx:
We didn't display this in the web UI, and we shouldn't have the API
loophole if we're serious about folks not being able to disappear on
their own.

See #621


mwillbanks:
I wanted to point out that self-deletion is possible by an administrator as an edge case, I'm not certain that we care about at this time but it is a possibility.

iansltx:
Corner case even if we closed the direct edge case: admin creates another account, promotes that other account to admin, deletes the original account. Superusers are slippery ;)

PR #624, opened 2019-02-09
WIP: Allow usage of ActionControllers

heiglandreas:
This commit allows to use action controllers that contain a single action that is executed when the __invoke-method is executed.


zghosts:
did you rerun composer update after the change?
I'm getting: 
"Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them."
composer update or composer update --lock should fix this

heiglandreas:
As this PR doesn't include any changes to composer.json or composer.lock: no, I didn't. Seemed not to be relevant...

iansltx:
Yep, that's pre-existing. We need to land #446 first.

zghosts:
whoops this remark was meant for #446

tests/routers/ActionControllerRouteTest.php
```diff
@@ -0,0 +1,173 @@
...
+     * @throws Exception
+     */
+    public function testDispatch(array $config, $controller, Request $request, $expectedException = false, $expectedExceptionCode = false)
+    {
+        $db = new PDO('sqlite:memory:');
+
+        $route = new ActionControllerRoute($controller);
+
+        try {
+            $this->assertEquals('val', $route->dispatch($request, $db, $config));
```
iansltx: I believe you need to fail here if you expected an exception and didn't get one?

tests/routers/ActionControllerRouteTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $request->expects($this->any())
+                ->method('getUrlElement')
+                ->with(1)
+                ->will($this->returnValue($urlElement));
+
+        return $request;
+    }
+}
+
+class TestController5
```
iansltx: Should these be properly namespaced/autoloaded rather than shoved into the same file?

heiglandreas:
Thanks for those comments. I was thinking also about injecting the Request into __invoke so thanks for. confirming that. A second thought that crossed my mind just yesterday was whether we shoumd try to implement PSR7/PSR15 here as well?

iansltx:
+1 for 7+15. Personal preference for Slim\Http request and response objects but I can live with Diactoros too.

heiglandreas:
Given that there was a personal intersection between Slim and joind.in I find slim/http a felicitous solution

svpernova09:
@heiglandreas What's the plan regarding @iansltx's review here?

svpernova09:

PR #625, opened 2019-02-10
Update composer.lock

ericpoe:
Update `composer.lock` to match `composer.json`

This resolves an error when running `scripts/update-repositories.sh` for `composer.lock`.


iansltx:
@ericpoe This will be resolved when #446 lands I believe, and adding this in now would cause #446 to conflict. Am I missing something?

ericpoe:
@iansltx You are correct. Closing.

PR #629, opened 2019-02-11
#628 Comment verification were throwing an exception in prod

mcneely:
Fixed Issue where comments were throwing an exception I suspect this fixes #628.

Also cleaned up code that was referencing a variable that did not exist.

PR #630, opened 2019-02-11
fixes the spamcheck on the contactform

zghosts:
Contactform submission is being blocked due to passing a string instead of an array to the spam check


zghosts:
duplicate to #629

mcneely:
Considering I had already told you I was working on this you're the one that should have "looked before you leaped".

PR #632, opened 2019-02-13
Implement pimple as a psr11 compatible di container

zghosts:
PR #629 has been merged into this branch as it is the testability of this PR which sparked the need to have a DI to improve testability.


zghosts:
Not quite sure about the container factory and the static container property therein, but it's a proof of concept, let's discuss it

src/controllers/ContactController.php
```diff
@@ -56,23 +96,19 @@ public function contact(Request $request, PDO $db)
...
 
         // run it by akismet if we have it
-        if (isset($this->config['akismet']['apiKey'], $this->config['akismet']['blog'])) {
-            $spamCheckService = new SpamCheckService(
-                $this->config['akismet']['apiKey'],
-                $this->config['akismet']['blog']
-            );
-            $isValid          = $spamCheckService->isCommentAcceptable(
-                $data['comment'],
+        if (isset($this->spamCheckService)) {
```
iansltx: Can we enforce this dependency instead, e.g. with a SpamCheckInterface and a dummy class that always returns true? That'll simplify code paths here.

src/routers/Route.php
```diff
@@ -105,16 +105,17 @@ public function setParams(array $params)
      *
      * @param Request $request The Request to process
      * @param PDO $db The Database object
-     * @param mixed $config The application configuration
+     * @param \Psr\Container\ContainerInterface $container The application configuration
      *
      * @return mixed
      */
-    public function dispatch(Request $request, $db, $config)
+    public function dispatch(Request $request, $db, $container)
```
iansltx: Add typehints here?

zghosts: Done

src/controllers/ContactController.php
zghosts: Added a NullSpamCheckService and a SpamCheckServiceInterface.
Good call!!

src/controllers/ContactController.php
```diff
@@ -32,20 +60,26 @@ public function contact(Request $request, PDO $db)
...
         $clientId         = $request->getParameter('client_id');
         $clientSecret     = $request->getParameter('client_secret');
-        $this->oauthModel = $request->getOauthModel($db);
-        if (! $this->oauthModel->isClientPermittedPasswordGrant($clientId, $clientSecret)) {
+        $oauthModel = $request->getOauthModel($db);
+        if (! $oauthModel->isClientPermittedPasswordGrant($clientId, $clientSecret)) {
             throw new Exception("This client cannot perform this action", 403);
         }
 
+        if (! isset($this->emailService)) {
```
iansltx: Don't need this anymore since we're constructor-injected :)

src/controllers/ContactController.php
```diff
@@ -55,24 +89,17 @@ public function contact(Request $request, PDO $db)
...
-            if (!$isValid) {
-                throw new Exception("Comment failed spam check", 400);
-            }
+        $isValid = $this->spamCheckService->isCommentAcceptable(
+            $data,
+            $request->getClientIP(),
+            $request->getClientUserAgent()
+        );
+
+        if (!$isValid) {
```
iansltx: Take it or leave it, but we don't need the temp var here.

src/controllers/ContactController.php
zghosts: done

src/controllers/ContactController.php
zghosts: and done

mcneely:
I wholeheartedly agree we need a DI container. My only big concern is that I suspect pimple, like Silex, is going to get abandoned.  With Symfony 4 being highly componentized it makes no sense for them to continue to maintain 2 separate DI containers. I'm just wondering if it might be in out best interest to go ahead and use the symfony di container instead.  I've put the question out to Fabien about this https://twitter.com/McNeely/status/1095680045325127680

@iansltx and @zghosts  On a side note it's probably a good idea to have a meeting as some point to discuss long term architectural goals for the project.  Having those set would go a long way in making proper short term decisions and preventing accrual of technical debt that prevents us from reaching that goal.

heiglandreas:
As pimple is PSR-11 compatible it is easy to exchange the DIC with something else should time arise.

As joind.in is not only a service but shall also be a project that is easy to get into I'd rather not use *huge* solutions but ones that allow people to actually see rather fast what's going on so that they can understand easily how things work together. Pimple seems to be such a solution. 

Additionally I'd also like to encourage us to see joind.in as a possibility to learn new stuff and check out new, smaller, probably unknown projects to also show to newcomers that there are ways beside the big frameworks.

Especially in this case, where changing the DIC should be rather easy (if not we f*cked up during the implementation) should the need arise we're rather safe with using pimple.

zghosts:
I think there are a lot of architectural choices to be made, and we should have a meeting on this.
with the current implementation the API is in no way coupled to a specific container implementation. The container itself is only injected into the router and as a PSR-11 ContainerInterface

Using Pimple now lets us achieve the shortterm goals of improving testability, which is lacking at many levels. The advantage of pimple is that it doesn't require the container to be compiled, integrating Symfony DI (which also has my preference) requires a lot more than integrating Pimple.

As swapping it out later for any other PSR-11 compatible container should be relatively easy, I would vote to merge this in order to have a more solid foundation to decouple the current code and improve testability and coverage, so it will be easier to migrate it to a different platform, without having to write a lot of boilerplate code in order to integrate SymfonyDI.

In future I'm guessing we'll be looking at converting the api to run om something like API-Platform, but that is a longterm architectural decision, we most certainly need to have a meeting on that topic.

iansltx:
As I mentioned earlier in the thread (and to agree with @heiglandreas and @zghosts), we can revisit the DIC flavor choice later, but Pimple is a solid choice (and one I've personally used extensively). It's also simple enough that, in the extremely unlikely event that we ended up heavily tied to it, we could maintain it ourselves (there isn't much to it).

Also the tweet thread you linked to seems to go exactly contrary to your fear that Pimple will get abandoned: Fabien said it's stable and not going away  Smaller projects that are feature-complete (and Pimple, for it's purposes, is) are easier to maintain when something does need changing, and having a dependency that doesn't move out from under us every so often is a benefit rather than a detriment. See also: Guzzle finally staying on a major version for awhile :)

We can certainly have architectural discussions that are longer-ranging at some point. But TBH that should be after both the host move and the flurry of commit activity we have now

PR #633, opened 2019-02-16
Added build scripts to composer.json.

mcneely:
Pointed travis and build.xml to composer scripts.
Tested both in Travis and Phing


michaelcullum:
To be honest, I'd rather split them out in the travis scripts as Travis is better when it knows what are different test suites.

Also, it's not really 'building' anything. It's running a series of test suites.

mcneely:
@michaelcullum This allows developers to run individualized tests from the command line not just the whole suite and establishes a single source of truth. I just used the build name because all those came from build.sh

michaelcullum:
Yep. I was referring to Travis, suggesting we explicitly split it out there.

If you want to have something run all too for development then +1 but I don't think it should be called build.

Also, I feel like this is a bit of an abuse of composer. This really all belongs in a makefile probably.

mcneely:
@michaelcullum I already renamed build.  Also using composer for scripts it not really a new thing.  Even Zend does it: https://github.com/zendframework/zend-expressive/blob/master/composer.json

mcneely:
@michaelcullum I see what you mean now. Putting the calls individually in Travis. My only concern would be that we still might want to do the full checks from the command line which would mean duplicating things in Travis and composer.  I'd rather not add a makefile to the mix, that just adds unneeded extra complexity.

composer.json
```diff
@@ -51,6 +51,24 @@
...
+        "diff": [
+            "git diff origin/master... -- > diff.txt",
+            "diffFilter --phpunit diff.txt build/logs/clover.xml 80"
+        ],
+        "check": [
+            "mkdir -p build/logs",
+            "@lint",
+            "@smell",
+            "@security",
+            "@unit",
```
zghosts: I guess @unit should be @test

composer.json
```diff
@@ -51,6 +51,24 @@
             "tests/views"
         ]
     },
+    "scripts": {
+        "test": "phpunit -c . tests/",
+        "lint": "parallel-lint --exclude vendor .",
+        "smell": "phpcs --standard=tools/codesniffer/JoindInPSR2/ruleset.xml --ignore=**/config.php,**/database.php,vendor,tools,tests/bootstrap.php --extensions=php --report-checkstyle=build/logs/checkstyle.xml --runtime-set ignore_warnings_on_exit true -p .",
```
zghosts: I feel sniff would be a better name for this task

composer.json
```diff
@@ -51,6 +51,24 @@
...
+        "security": "security-checker security:check composer.lock",
+        "diff": [
```
zghosts: I would call this task coverage instead of diff

zghosts:
Otherwise I kind of like the idea of configuring the build tasks in composer.json.
I get @michaelcullum's remark that it's more of Makefile thing, however it's convenient to run the tests individually and it doesn't require any additional tools to be installed in order to run the test and qa tools.

That said, I think @michaelcullum does have a point regarding the travis configuration, but that can be changed at a later stage, as their not split out currently anyway.

composer.json
```diff
@@ -63,9 +63,9 @@
         "check": [
             "mkdir -p build/logs",
             "@lint",
-            "@smell",
+            "@snill",
```
zghosts: typo ;)

composer.json
mcneely: maybe filter?   The thing is it's not generating coverage, phpunit is already doing that via the dist file.

composer.json
mcneely: oops

composer.json
mcneely: done.

composer.json
mcneely: oops.

zghosts:
I still have an issue with the "diff" task as your not creating a diff, rather your checking the testcoverage of the changed code, based on the diff against master.

PR #638, opened 2019-03-19
Use current version of Guzzle to interact with OAuth

iansltx:
Requires version bump on OAuth subscriber

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,8 +66,7 @@
             "@lint",
             "@sniff",
             "@security",
-            "@test",
-            "@coverage"
+            "@test"
         ]
     },
     "config" : {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -66,15 +66,11 @@
...
             "@security",
-            "@test",
-            "@coverage"
+            "@test"
         ]
     },
     "config" : {
         "preferred-install": "dist",
-        "platform" : {
-            "php": "5.6.6"
```
ericpoe: When we upgrade joindin-vm to PHP 7.3, I'd like to see this replaced with `"php": "7.3.1"` just so we don't get caught unawares when PHP 7.4 rolls around and we merge in a PR that requires 7.4.

mcneely: We'll actually already on 7.3, hence the twitter problem.

PR #646, opened 2019-03-26
Psr4

zghosts:
I realize this is a huge PR, I might need to split it up into several PR's.
It moves all classes into a PSR-4 namespace hierarchy.
The structure of the namespaces might require some discussion..

additionally
- it moves package.json from tests/frisby to the root of the application
- it add output buffering to the ApiView to stop the "headers have already been sent" warning.


src/Controller/FacebookController.php
```diff
@@ -91,7 +96,7 @@ public function logUserIn(Request $request, PDO $db)
         }
 
         trigger_error('Unexpected Facebook error (' . $res->getStatusCode()
-            . ': ' . $res->getBody() . ')', E_USER_WARNING);
+                      . ': ' . $res->getBody() . ')', E_USER_WARNING);
```
ericpoe: This block is really difficult to read. Can we clean it up using something like:

```php
trigger_error(
    sprintf(
        'Unexpected Facebook error (%s: %s)',
        $res->getStatusCode(),
        $res->getBody()
    ),
    E_USER_WARNING
);
```

src/Controller/TalksController.php
```diff
@@ -34,7 +55,7 @@ public function getTalkStarred(Request $request, PDO $db)
     {
         $this->setDbAndRequest($db, $request);
         $talk_id = $this->getItemId($request);
-        $mapper = $this->getMapper('talk');
+        $mapper  = $this->getMapper('talk');
```
ericpoe: What is the benefit to aligning assignment operators? If we were to add a `$longerName` var on the next line, then that would require updating lines 57 & 58 also, and that would make tracing through the git history more difficult.

tests/ContainerFactoryTest.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @test
+     * @covers \Joindin\Api\ContainerFactory::build
+     */
+    public function spamCheckServiceIsNullCheckerWhenDisabled()
+    {
+        $container = ContainerFactory::build([], true);
+        $this->assertTrue($container->has(\Joindin\Api\Service\SpamCheckServiceInterface::class));
```
ericpoe: Now that we're using namespaces, the `\` prefix means global namespace, so I believe this should just be 

```php
$this->assertTrue($container->has(Joindin\Api\Service\SpamCheckServiceInterface::class));
```

Better yet would be to add a `use Joindin\Api\Service\SpanCheckServiceInterface` line to the top of this file and just have this line be 

```php
$this->assertTrue($container->has(SpamCheckServiceInterface::class));
```

tests/RequestTest.php
```diff
@@ -1,18 +1,17 @@
...
-
 class RequestTest extends TestCase
 {
     /**
      * Make sure we have everything we need - in this case the config
      */
     public function setUp()
     {
-        include __DIR__ . '/../../src/config.php';
+        include __DIR__ . '/../src/config.php';
```
ericpoe: `dirname()` would give you the parent path and `dirname(__DIR__)` would give you the absolute path of the parent directory of the current file's directory. So perhaps this might be a little cleaner:

```suggestion
        include dirname(__DIR__) . 'src/config.php';
```

ericpoe:
I see at least 3 PRs here:

- Migration to PSR-2
- Frisby code-style changes
- PHP code-style changes

Please separate these changes out into their own PRs. I would prefer to see focus on the PSR-2 migration PR first because I'm working on a PR that makes changes to some of these newly renamed files and their included namespaces. I would imagine that others might be doing similar effort.

### PSR-2 Migration Changes - comments

I ❤️ it! I've tried this at least twice in the last few years and never submitted a PR because I couldn't quite get it to work.

Now that we're using namespaces, all `Joindin\\` references should not have a `\` prefix because that means global namespace.

ex.
- `\Joindin\*` --> `Joindin\*`

Now, all global classes should have the `\` prefix.

ex.
- `Exception` -->` \Exception`
- `PDO` -->` \PDO`
- `DateTime` --> `\DateTime`

Once the `use` statements that have global objects are prefixed with `\`, please alphabetize the use statements (PHPStorm: Code > Optimize Imports).

### Frisby changes - comments & questions

- Tabs --> spaces FTW!
- `package.json` move, yay!
- Should we have a committed `package-lock.json` to go along with `package.json` to ensure that everyone is using the right versions of the various libraries?

### PHP code-style changes - comments & questions

Most of these seem to be whitespace changes.

- What is the benefit to aligning code to operators (`=`, `=>`, etc)? The problem that I see with this is that adding a new line with the same operator would require editing all lines that are aligned with that operator, which adds to a more complex PR and muddles the git history of that line. Also, this extends the length of the line, potentially beyond the 120 spaces suggested by PSR-2, and adds to eyes scanning further left and right and can require having to scroll left and right in order to read.
- Removing the space after a typecast (ex. `(string) $this->getValue()` --> `(string)$this->getValue()` seems to be a personal preference. What is the reason for this? If this is something that we should continue to do henceforth, then we should [add it to the `ruleset.xml` ](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingspaceaftercast) file for CodeSniffer. I would prefer to see a space after the cast.
- Removing the space after a not-operator (ex. `! $thing` --> `!$thing`) seems to be a personal preference. What is the reason for this? If this is something that we should continue to do henceforth, then we should [add it to the `ruleset.xml` ](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties#genericformattingspaceafternot) file for CodeSniffer. I would prefer to see no space after the cast.
- Basically, I would prefer to see CodeSniffer rules for each whitespace violation that we don't like. Otherwise, it should be allowed.

tests/ContainerFactoryTest.php
zghosts: Agreed.

src/Controller/FacebookController.php
zghosts: changed in latest commits

tests/RequestTest.php
zghosts: I'd much rather turn config.php into a configuration class, which would be even cleaner, however as this pr is big enough as it is I would like tot do those optimisations after this one

src/Model/EventMapper.php
```diff
@@ -104,15 +110,15 @@ public function getEventById($event_id, $verbose = false, $activeEventsOnly = tr
      * @param bool $activeEventsOnly
      * @param bool $transform
      *
-     * @return false|bool the event detail
+     * @return bool|array the event detail
```
ericpoe: Since this never returns `true`, `@return false|array` would be more accurate here.

src/Router/DefaultRouter.php
```diff
@@ -11,6 +15,6 @@ class DefaultRouter extends BaseRouter
      */
     public function getRoute(Request $request)
     {
-        return new Route('DefaultController', 'handle');
+        return new Route('Joindin\Api\Controller\DefaultController', 'handle');
```
ericpoe: Could this be `return new Route(DefaultController::class, 'handle');` instead and import the path as a `use` statement instead?

zghosts: That would be better, will do in a followup PR

src/Model/EventMapper.php
zghosts: True, will check in a followup PR

PR #653, opened 2019-04-20
Apply fixes from StyleCI

svpernova09:
This pull request applies code style fixes from an analysis carried out by [StyleCI](https://github.styleci.io).

---

For more information, click [here](https://github.styleci.io/analyses/86bDE4).


svpernova09:
Here's a PR we can review to see what changes Style-CI will have on the code base. If we get approvals we can move forward.

iansltx:
Looks like StyleCI is venturing well into "matters of opinion" territory here, with variable alignment, use of single quotes, and (the biggest nonstarter for me due to its effect on legibility) dropping spaces on string concatenation. If we're going to have a tool affect code like this, let's make sure we agree with the changes it's doing first, as they're obviously beyond PSR-2 (or PSR-12 for that matter).

In any case (hence closing the PR I suspect) we'll want to "true up" this _after_ we have big PRs (like PSR-4) merged, otherwise we've likely created a sea of conflicts that would've been otherwise easily avoided.

svpernova09:
> Looks like StyleCI is venturing well into "matters of opinion" territory here, with variable alignment, use of single quotes, and (the biggest nonstarter for me due to its effect on legibility) dropping spaces on string concatenation. If we're going to have a tool affect code like this, let's make sure we agree with the changes it's doing first, as they're obviously beyond PSR-2 (or PSR-12 for that matter).

As discussed in https://github.com/joindin/joindin-web2/issues/749 the plan is to enforce PSR-2. Anything past PSR-2 should be decided on in favor of readability IMHO. I don't plan on dying on any code style hills. I just want to get the automation in place.

> In any case (hence closing the PR I suspect) we'll want to "true up" this after we have big PRs (like PSR-4) merged, otherwise we've likely created a sea of conflicts that would've been otherwise easily avoided.

I'm fine waiting so long as it's reasonably time boxed and agree it'd be better to get larger PRs merged. Between now and then we should work through a Style-CI configuration that offers PSR-2 support (and we can jump to PSR-12 when it's finalized)

iansltx:
Sounds reasonable. I have (previously documented) opinions against StyleCI's approach, but if we can get a ruleset that trues up to PSR-2 and otherwise doesn't do much to the codebase when first applied I'll sign off on it.

Re: outstanding PRs, if they're still outstanding by this time next week, ping me and I'll do my darnedest to review if that's what they're waiting on.

Issue #617, opened 2019-02-03
Update tests for compatibility with current PHPUnit versions


iansltx:
Will need to look at backward/forward compatibility here. The tests work with 5.x but they're using the old non-namespaced syntax. PHPUnit 5.x is the latest PHP 5.6 compatible version of PHPUnit so this upgrade may be a multi-step process depending on functionality deprecated/removed since PHPUnit 5.

heiglandreas:
This is fixed with #654

PR #654, opened 2019-05-02
Upgrade testing suite

ericpoe:
This updates a few libraries that we use for testing in the CI. This also updates testing documentation since we are not using phing in the VM.

This takes care of #617 

I would prefer to fix the style-ci "errors" via php_codesniffer rules in a separate PR since that will involve adding new rules to our `phpcs.xml.dist` config file.


tests/controllers/EventHostsControllerTest.php
```diff
@@ -2,18 +2,16 @@
 
 namespace JoindinTest\Controller;
 
-use Mockery as M;
 use PHPUnit\Framework\TestCase;
 
 class EventHostsControllerTest extends TestCase
 {
-    /**
-     * @expectedException \Exception
```
ericpoe: annotation-level exception assertions are deprecated in PHPUnit 8. https://github.com/sebastianbergmann/phpunit/issues/3332

ericpoe:
@zghosts based on most of our tests, I'd suggest that we use the `test` prefix rather than the `@test` annotation.

PR #657, opened 2019-05-06
Fix unstar-talk route mapping

iansltx:
Fixes https://github.com/joindin/joindin-web2/issues/813

Changes API response code (which is fine since the route didn't work
anyway) to 205 to be more semantically correct

### DIFF
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -201,9 +201,9 @@
         "verbs": ["DELETE"]
     },
     {
-        "path": "/talks/[\\d+]]/starred?",
+        "path": "/talks/[\\d]+/starred/?$",
         "controller": "Joindin\\Api\\Controller\\TalksController",
-        "action": "deleteAction",
+        "action": "deleteTalkStarred",
         "verbs": ["DELETE"]
     },
     {
--- a/src/config/routes/2.1.json
+++ b/src/config/routes/2.1.json
@@ -203,7 +203,7 @@
     {
         "path": "/talks/[\\d]+/starred/?$",
         "controller": "Joindin\\Api\\Controller\\TalksController",
-        "action": "deleteTalkStarred",
+        "action": "removeStarFromTalk",
         "verbs": ["DELETE"]
     },
     {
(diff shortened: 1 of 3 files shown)


src/config/routes/2.1.json
```diff
@@ -201,9 +201,9 @@
         "verbs": ["DELETE"]
     },
     {
-        "path": "/talks/[\\d+]]/starred?",
+        "path": "/talks/[\\d]+/starred/?$",
         "controller": "TalksController",
-        "action": "deleteAction",
+        "action": "deleteTalkStarred",
```
heiglandreas: Could we rename that function to `unstarTalk`? This `deleteTalkStarred` sounds more like "delete all talks that are starred" (which is – hopefully – not what we do)

iansltx: Storm gave me a squiggly underline on "unstar" so I matched "removeXfromY" seen in at least one other route. lmk what you think

svpernova09:
There was 1 failure:
1) JoindinTest\Controller\TalksControllerDeleteTest::removeStarFromTalkFailsWhenNotLoggedIn
Failed asserting that exception of type "Error" matches expected exception "\Exception". Message was: "Call to undefined method TalksController::deleteTalkStarred()" at
/home/travis/build/joindin/joindin-api/tests/controllers/TalksControllerDeleteTest.php:31

.styleci.yml
```diff
@@ -0,0 +1,7 @@
+preset: psr2
+
+risky: false
+
+finder:
+  name:
```
mcneely: This does not exclude vendor directories. you really don't want this hitting packages.

iansltx: Does StyleCI run composer install prior to running its analysis? We have vendor gitignored so this seems like it would be a nice bit of insurance against accidentally checking in vendor.

iansltx:
@svpernova09 thx for pointing that bit out; should be fixed now

.styleci.yml
mcneely: Either way you shouldn't include the file until we decide what is going in it. And this is ancillary to the bug fix.

mcneely:
Woah, someone did something wrong.

iansltx:
GitHub's merge-to-catch-up be cray, yo

.styleci.yml
iansltx: Rebased it out.

PR #659, opened 2019-05-15
Add Transliteration for meaningful slugs

heiglandreas:
This adds a transliteration step to create meaningful slugs even when
the title is written in non-latin script

Fix joindin/joindin-web2#815


dstockto:
Does the intl extension exist on the production server currently?

heiglandreas:
According to https://docs.platform.sh/languages/php/extensions.html the intl extension is available and enabled by default.

lisachenko:
Thank you, @heiglandreas  👍

PR #660, opened 2019-05-16
Added two tests for access token generation.

DanielCraigie:


iansltx:
@DanielCraigie Tokens should be ASCII-compatible, so I believe we actually want non-multibyte strlen()s here.

Pushing that change will also re-trigger StyleCI. I dropped alignment rules just now in the config (which isn't in-repo at the moment) so it should pass now. Didn't make sense to keep those rules since StyleCI will flag based on the entire file when you didn't edit anything else, and thus blow up the diff.

heiglandreas:
It does make sense to check using mb_strlen as the DB-field is a varchar(16) in a UTF8 table. So it actually can hold multibyte characters. Though I see it like @iansltx that that doesn't make sense. So we perhaps should assert before setting the access token that it actually is only ASCII based alphanumeric....

iansltx:
Checked the impl and we're handing back 16 hexidecimal characters at this point, so "it should be ASCII" seems perfectly reasonable to assert.

Though this brings up why we're still using `sha1()` for anything at this point, given that it's getting cheaper by the minute to so SHA1 collisions.

PR #667, opened 2019-05-30
add robots.txt disallowing all bots

zghosts:


iansltx:
I was wondering whether there was a secret where an empty robots file was equivalent to disallow all heh

PR #668, opened 2019-06-02
DB patch to fix oauth token generation error when access_token_secret…

DanielCraigie:
… is not used.

Default DB schema requires an access_token_secret but this is not generated in the current version.

Users trying to login see the page refresh with the login credential section still visible. The API login request returns a 500 that is generated by the DB insert command to the oauth_access_tokens table.

This patch removes the NOT NULL constraint from the column allowing records to be inserted.

I have also added a line to the API README file to help newcomers apply their patches.


svpernova09:
Thanks!

PR #672, opened 2019-06-07
Tag routes with controller/action info for New Relic if it's installed

### LINKED ISSUE #671: When New Relic is installed, tag HTTP requests with controller + action names
Right now everything in New Relic drops into the /index.php bin for API endpoints. That means any metrics we're collecting are painted with a broad brush, rather than providing insight into particularly heavy endpoints. We can use newrelic_name_transaction() (when it exists) to provide controller/action info to make those metrics a bit more useful.

iansltx:
Fixes #671

### DIFF
--- a/src/Router/Route.php
+++ b/src/Router/Route.php
@@ -121,13 +121,19 @@ class Route
         $className = $this->getController();
         $method    = $this->getAction();
 
-        if ($container->has($className)) {
-            $controller = $container->get($className);
-            if (method_exists($controller, $method)) {
-                return $controller->$method($request, $db);
-            }
+        if (!$container->has($className)) {
+            throw new RuntimeException('Unknown controller ' . $request->url_elements[2], 400);
+        }
+
+        if (!method_exists($controller = $container->get($className), $method)) {
             throw new RuntimeException('Action not found', 500);
         }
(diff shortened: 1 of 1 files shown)


iansltx:
Yeah, threw it behind a function_exists call to avoid a hard dependency. Easy enough to revert this if we drop 'em.

src/Router/Route.php
```diff
@@ -124,6 +124,11 @@ public function dispatch(Request $request, $db, ContainerInterface $container)
         if ($container->has($className)) {
             $controller = $container->get($className);
             if (method_exists($controller, $method)) {
+                if (function_exists('newrelic_name_transaction')) {
```
heiglandreas: These are too many levels of indentation.

Can we remove them like this?

```php
if (! $container->has($clsasName)) {
    throw new RuntimeException('Unknown controller ' . $request->url_elements[2], 400);
}

$controller = $container->get($className);
if (! method_exists($controller, $method)) {
    throw new RuntimeException('Action not found', 500);
}
if (function_exists('newrelic_name'transaction')) {
    $controllerWithoutNamespace = @end(explode('\\', $controller));
    newrelic_name_transaction($controllerWithoutNamespace . '.' . $method);
}

return $controller->$method($request, $db);
```

iansltx:
@heiglandreas lmk if this works

PR #675, opened 2019-06-07
WIP: Add POST /users/:id/admin endpoint (works similar to /trusted)

heiglandreas:
Obsoletes https://github.com/joindin/joindin-api/pull/620

### DIFF
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -438,23 +438,68 @@ class UsersController extends BaseApiController
      */
     public function setTrusted(Request $request, PDO $db)
     {
+        $this->setAdminOnlyBooleanField($request, $db, 'trusted');
+    }
+
+    /**
+     * Allow users to be set as admin
+     *
+     * @param $request Request
+     * @param $db      PDO
+     *
+     * @throws Exception
+     */
+    public function setAdmin(Request $request, PDO $db)
+    {
+        $this->setAdminOnlyBooleanField($request, $db, 'admin');
+    }
+
+    /**
+     * Handles setting a boolean value that only an admin can set, since we have the same pattern happening 2x
+     * (and may add setting for the activity bool as well)
+     *
+     * @param $request Request
+     * @param $db      PDO
+     * @param $field   string
+     *
+     * @throws Exception
+     */
+    protected function setAdminOnlyBooleanField(Request $request, PDO $db, $field)
+    {
         if (false == ($request->getUserId())) {
             throw new Exception("You must be logged in to change a user account", 401);
         }
 
+        if (!in_array($field, ['trusted', 'admin'])) {
+            throw new \InvalidArgumentException('Unknown boolean field change attempted', 501);
+        }
+
         $user_mapper = $this->getUserMapper($db, $request);
         if (!$user_mapper->isSiteAdmin($request->getUserId())) {
-            throw new Exception("You must be an admin to change a user's trusted state", 403);
+            throw new Exception("You must be an admin to change a user's $field state", 403);
         }
 
         $userId = $this->getItemId($request);
-        if (!is_bool($trustedStatus = $request->getParameter("trusted", null))) {
-            throw new Exception("You must provide a trusted state", 400);
+        if (!is_bool($fieldStatus = $request->getParameter($field, null))) {
+            $article = $field === 'admin' ? 'an' : 'a';
+            throw new Exception("You must provide $article $field state", 400);
         }
 
-        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
+        switch ($field) {
+            case 'trusted':
+                $wasSuccessful = $user_mapper->setTrustedStatus($fieldStatus, $userId);
+                break;
+            case 'admin':
+                $wasSuccessful = $user_mapper->setAdminStatus($fieldStatus, $userId);
+                break;
+            default:
+                throw new \InvalidArgumentException("Unknown boolean field change attempted: $field", 501);
+        }
+
+        if (!$wasSuccessful) {
             throw new Exception("Unable to update status", 500);
         }
+
         $view = $request->getView();
         $view->setHeader('Content-Length', 0);
         $view->setResponseCode(204);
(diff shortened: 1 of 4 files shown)


src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
      */
     public function setTrusted(Request $request, PDO $db)
+    {
+        $this->setAdminOnlyBooleanField($request, $db, 'trusted');
+    }
+
+    /**
+     * Allow users to be set as admin
+     *
+     * @param $request Request
```
BackEndTea: These 2 `@param` annotations don't really add anything.

This comment can probably be removed completely.

src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
         }
 
+        if (!in_array($field, ['trusted', 'admin'])) {
+            throw new \InvalidArgumentException('Unknown boolean field change attempted', 501);
+        }
+
         $user_mapper = $this->getUserMapper($db, $request);
         if (!$user_mapper->isSiteAdmin($request->getUserId())) {
-            throw new Exception("You must be an admin to change a user's trusted state", 403);
+            throw new Exception("You must be an admin to change a user's $field state", 403);
```
BackEndTea: Probably better to use `sprintf` over string interpolation. It keeps it more readable.

src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
+
+    /**
+     * Allow users to be set as admin
+     *
+     * @param $request Request
+     * @param $db      PDO
+     *
+     * @throws Exception
+     */
+    public function setAdmin(Request $request, PDO $db)
```
BackEndTea: Since the code requires php 7.3, this method can declare a void return type
```suggestion
    public function setAdmin(Request $request, PDO $db): void
```

src/Controller/UsersController.php
iansltx: Fair enough. Just checked and we're using sprintf in plenty of other places here

svpernova09: IMHO 

`"You must be an admin to change a user's $field state"`

Reads easier than 

`sprintf("You must be an admin to change a user's %s  state", $field);`

src/Controller/UsersController.php
iansltx: Storm complains about uncaught, un-annotated exceptions if I drop the docblocks. Just pushed a commit that suppresses these complaints, which allows for dropping docblocks without highlighting the sole method call here.

src/Controller/UsersController.php
iansltx: Guess we need a style convention around this then. It's easier to search the codebase for sprintf than string interpolation, so I know the former exists a fair amount, but not sure about the latter. My personal pref is for interpolation because this is PHP rather than C and we're dealing with strings all the way around rather than e.g. formatting floats, but I'd rather codify one way or another than have to decide whether to revert the commit I just pushed :)

src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
+     * Handles setting a boolean value that only an admin can set, since we have the same pattern happening 2x
+     * (and may add setting for the activity bool as well)
+     *
+     * @param $request Request
+     * @param $db      PDO
+     * @param $field   string
+     *
+     * @throws Exception
+     */
+    protected function setAdminOnlyBooleanField(Request $request, PDO $db, $field)
```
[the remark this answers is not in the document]
svpernova09: I'd rather see new language features (return types, scalar types) added via PR across the entire code base as opposed to slowly PR by PR.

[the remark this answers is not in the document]
BackEndTea: Of course that would be the best, but we can already start doing this for new content.

iansltx:
@heiglandreas I've updated my PR with some StyleCI fixes, and rebased it against current master. May be worth switching back to it?

iansltx:
Specifically https://github.com/joindin/joindin-api/compare/master...iansltx:add-set-admin-endpoint

src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
-        if (!$user_mapper->setTrustedStatus($trustedStatus, $userId)) {
+        switch ($field) {
+            case 'trusted':
+                $wasSuccessful = $user_mapper->setTrustedStatus($fieldStatus, $userId);
+                break;
+            case 'admin':
+                $wasSuccessful = $user_mapper->setAdminStatus($fieldStatus, $userId);
+                break;
+            default:
+                throw new \InvalidArgumentException("Unknown boolean field change attempted: $field", 501);
```
BackEndTea: This logic is duplicated here and on line 474.

tests/Controller/UsersControllerTest.php
```diff
@@ -264,6 +264,27 @@ public function testSetTrustedWithNoUserIdThrowsException()
         $usersController->setTrusted($request, $db);
     }
 
+    /**
+     * Ensures that if the setAdmin method is called and no user_id is set,
+     * an exception is thrown
+     *
+     * @return void
+     *
+     * @test
```
BackEndTea: `@test` annotation isn't needed if the `test` prefix is used

tests/Controller/UsersControllerTest.php
```diff
@@ -295,7 +316,38 @@ public function testSetTrustedWithNonAdminIdThrowsException()
...
+            ->getMock();
+
+        $userMapper
+            ->expects($this->once())
+            ->method('isSiteAdmin')
+            ->will($this->returnValue(false));
+
+        $usersController->setUserMapper($userMapper);
 
+        self::expectException(\Exception::class);
```
BackEndTea: Every Exception is an `Exception`. It may be beneficial to use more specific error classes

src/Controller/UsersController.php
```diff
@@ -437,24 +437,69 @@ public function deleteUser(Request $request, PDO $db)
...
+     *
+     * @throws Exception
+     */
+    protected function setAdminOnlyBooleanField(Request $request, PDO $db, $field)
     {
         if (false == ($request->getUserId())) {
             throw new Exception("You must be logged in to change a user account", 401);
         }
 
+        if (!in_array($field, ['trusted', 'admin'])) {
```
BackEndTea: If we use `true` as a third variable for `in_array`, then we can be sure about the types here.

svpernova09:
@heiglandreas Can you take a look at @BackEndTea's comments and resolve conflicts here please.

svpernova09:

PR #676, opened 2019-06-07
fixes some ci issues

zghosts:


iansltx:
@zghosts will review as soon as it's current with master again

PR #677, opened 2019-06-07
Database performance improvements (indexes, query simplification)

fatalmind:
Some suggestios to improve performane #DPC19

### DIFF
--- a/db/patch78.sql
+++ b/db/patch78.sql
@@ -1,0 +1,62 @@
+-- Performance improvements: indexing and simplifiying query
+
+-- TABLE: talk_comments
+-- Rationale:
+-- * It is common to include the "active=1 and private = 0" filters
+-- * (e.g. to count number of comments per talk).
+-- rating and user_id are included to allow an index only scan (e.g. in get_talk_rating)
+CREATE INDEX idx_talk_priv_act ON talk_comments (talk_id, private, active, rating, user_id);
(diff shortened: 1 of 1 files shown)


iansltx:
@fatalmind You may want to base this branch on the one mentioned in https://github.com/joindin/joindin-api/pull/674, as I was working on something similar elsewhere earlier. You'll also need to bump the patch count so ours can both land without conflicting.

iansltx:
Also, given that get_talk_rating() is used 2x in the entire codebase, what do you think about replacing it with a view? Writing up the query that would result in that view now.

iansltx:
Something like this:

```
select round(avg(rating)), tc.talk_id
from talk_comments tc
left join talk_speaker ts on ts.talk_id = tc.talk_id
where tc.user_id != 0 && (ts.speaker_id IS NULL || tc.user_id != ts.speaker_id) &&
tc.rating != 0 && tc.private = FALSE && tc.active = TRUE
GROUP BY tc.talk_id;
```

iansltx:
Seems like your idx_talk_priv_act works nicely with that view too.

iansltx:
If you're up for the view changes, that'd be amazing. If not, this looks fine as-is, and I'll add a follow-on issue to deal with this function.

heiglandreas:
Obsoleted by #681

PR #680, opened 2019-06-07
check if $return_data is an array before enriching the metadata

### LINKED ISSUE #679: Fatal error when the result_data from the controller is not an array
When the result data from a controller is not an array, updating the metadata fails with:
Warning: Cannot use a scalar value as an array in /var/www/html/src/public/index.php on line 81

zghosts:
fixes #679


src/public/index.php
```diff
@@ -77,11 +77,9 @@ function handle_exception(\Throwable $e)
 
 $route = $router->getRoute($request);
 $return_data = $route->dispatch($request, $ji_db, $container);
-
-if ($return_data && isset($request->user_id)) {
+if (is_array($return_data) && isset($request->user_id)) {
```
heiglandreas: It's definitely better than before, but shouldn't we also check whether the array-key is set?

```php
if (isset($return['meta']['user_uri']) && isset($request->user_id)) {
```

heiglandreas: stupid me. Checking whether it's an array is perfeclty sufficcient sa the key is created if it doesn't exist.... :man_facepalming:

PR #682, opened 2019-06-09
Convert to PHP arrays for routing config

### LINKED ISSUE #637: Move from JSON to PHP array for route declaration
Yes, it's a minuscule performance improvement for what we're doing, but PHP arrays can be cached in opcache (and since the routes table is static it'll be cached once across all requests), while JSON blobs have to be read from disk every time. Would be easy enough to swap things over and we don't use the routes file in its JSON form anywhere else from what I can see.

dstockto:
Fixes route for creating comments on events
Closes #637


src/config/routes/2.1.php
```diff
@@ -0,0 +1,678 @@
...
+        'controller' => EventsController::class,
+        'action'     => 'rejectAction',
+        'verbs'      =>
+            [
+                'DELETE',
+            ],
+    ],
+    [
+        'path'       => '/events(/[^/]+)*/comments',
+        'controller' => EventCommentsController::class,
```
dstockto: This is corrected from the json route file.

src/config/routes/2.1.json
```diff
@@ -1,458 +0,0 @@
...
-    },
-    {
-        "path": "/events/(\\d+)/approval$",
-        "controller": "Joindin\\Api\\Controller\\EventsController",
-        "action": "rejectAction",
-        "verbs": ["DELETE"]
-    },
-    {
-        "path": "/events(/[^/]+)*/comments",
-        "controller": "Joindin\\Api\\Controller\\Event_commentsController",
```
dstockto: This is the broken part of the old config.

ericpoe:
@dstockto have you had a chance to look into that merge conflict?

dstockto:
@ericpoe I'll take a look - it was not conflicting when it was PR'd 3 weeks ago. I don't think github notifies when it starts conflicting.

zghosts:
@dstockto You might want to incorporate #684 into this PR

dstockto:
Agreed, @zghosts  - I think that's the cause of the conflict.

dstockto:
@mcneely @ericpoe @zghosts - This is no longer conflicting and includes the changes to routing from #684

PR #683, opened 2019-06-11
Don't use nonexistent method to pull JSON from Twitter response body

iansltx:
Noticed this in errors on New Relic. So user creation via Twitter SSO...or sign-in via Twitter SSO at all...is probably broken at the moment.


heiglandreas:
We should redo the whole social Auth thing using a third party library. And then modify the whole thing so that one user account can have multiple logins. And that creating a user-account via social login is a fast one-step thing

iansltx:
Yep...but might as well get this quick fix in first :)

PR #684, opened 2019-06-25
fix talk deletion route

zghosts:
fixes the talk delete route


iansltx:
Shouldn't this be the same pattern as what's used for getAction, namely:

```
        "path": "/talks/[\\d]+/?$",
```

?

zghosts:
That also works :)

PR #686, opened 2019-08-02
Don't fail when setting user attendance on an already-attending user

iansltx:
This makes this call idempotent, so duplicate calls won't cause 500s.


svpernova09:
Thanks!

PR #690, opened 2019-08-21
Enhancement: Add .editorconfig

localheinz:
This PR

* [x] adds a basic `.editorconfig`

💁‍♂️ For reference, see https://editorconfig.org/.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.json]
+indent_size = 2
```
localheinz: This currently applies to 2 out of 3 `.json` files. Will follow up with an adjustment to `composer.json`, for consistency, if there are no objections.

PR #691, opened 2019-08-21
Enhancement: Install dependencies in install section

localheinz:
This PR

* [x] installs dependencies with `composer` on Travis in `install` section 

💁‍♂ For reference, see https://docs.travis-ci.com/user/job-lifecycle/.

PR #692, opened 2019-08-21
Fix: Lock file is not up-to-date

localheinz:
This PR

* [x] validates `composer.json` and `composer.lock` on Travis in `before_install` section
* [x] updates `composer.lock`

💁‍♂ For reference, see https://getcomposer.org/doc/03-cli.md#validate.

❗️ Should probably be merged after #691, so that we keep sections in order.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ matrix:
     - php: 7.3
   fast_finish: true
 
+before_install:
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks!

localheinz:
Thank you, @svpernova09!

PR #694, opened 2019-08-21
Enhancement: Reference phpunit.xsd as installed with composer

localheinz:
This PR

* [x] fixes the wrapping in `phpunit.xml.dist`
* [x] references `phpunit.xsd` as installed with `composer`
* [x] keep attributes sorted by name

💁‍♂ Helps with auto-completion and validation in IDEs.


localheinz:
Thank you, @svpernova09!

PR #696, opened 2019-08-21
Enhancement: Add basic configuration for Dependabot

localheinz:
This PR

* [x] adds a configuration for [Dependabot](https://dependabot.com)

💁‍♂ Now it would only need to be enabled for this repository!

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,9 @@
+# https://dependabot.com/docs/config-file/
+
+version: 1
+
+update_configs:
+  directory: "/"
+  package_manager: "php:composer"
+  update_schedule: "live"
+  version_requirement_updates: "increase_versions"


svpernova09:
Thoughts on this @joindin/leadership-team?

mwillbanks:
Personally, I really like this.  I think the `version_requirement_updates` is good for right now (`increase_versions`), however, I do think it would be good to eventually get it to `widen_ranges` so that we move forward with larger version increases but that is easier when there is a develop branch.

Further, we should look at the `allowed_updates` configuration option a bit deeper; in the more general case the master branch is set for `security` only while the develop branch contains all dependencies.

Either way, it looks good to me!

localheinz:
Thank you, @mcneely, @mwillbanks, and @svpernova09!

localheinz:
@mcneely @svpernova09 

Has an integration been set up yet?

localheinz:
Sounds good, @svpernova09!

PR #705, opened 2019-10-03
Add rate limit user comments

### LINKED ISSUE #648: Add rate limiting for user comments
We just had a spammer hit us with >2k comments, in what appear to be rapid succession (will have to look at API logs because, to clean things up quickly, I nuked all of their comments).

I'd like to propose a rate limit that disallows (returns a 429 with a "please wait X more seconds") a comment if we've seen another comment from the same user in the last 20 seconds. We should include the wait time in both the error message (so the web UI can pick it up without modification) and the API (Retry-after header). This should provide somewhat of a foil against the kind of comment spam we just saw, which occurred through the website (vs. directly via the API).

On top of this, I'd also suggest a rolling rate limit of 30 comments per hour. We can analyze the current list of comments to see whether we'd catch any legitimate user via that, but my gut says that we wouldn't, since that number would correspond to a user rating every talk they've seen at a long/action-packed conference in a single sitting.

arjank:
Add rate limit for talk comments

Includes db migrations and some cleanup. Rate limit checks for 30 comments in 60 minutes (both configurable). If you try to add more, http status 429 is returned along with a message to wait.

Closes #648

### DIFF
--- a/src/Model/ApiMapper.php
+++ b/src/Model/ApiMapper.php
@@ -71,7 +71,8 @@ class ApiMapper
                     } else {
                         $tz = new DateTimeZone('UTC');
                     }
-                    $entry[$key] = (new DateTime('@' . $row[$value]))->setTimezone($tz)->format('c');
+                    $time = is_numeric($row[$value]) ? '@' . $row[$value] : $row[$value];
+                    $entry[$key] = (new \DateTimeImmutable($time))->setTimezone($tz)->format('c');
                 } else {
                     if (array_key_exists($value, $row)) {
                         $entry[$key] = $row[$value];
--- a/src/Model/TalkCommentMapper.php
+++ b/src/Model/TalkCommentMapper.php
@@ -246,7 +246,7 @@ class TalkCommentMapper extends ApiMapper
      * @param int $user_id
      * @throws RateLimitExceededException
      */
-    public function checkRateLimit(int $user_id)
+    public function checkRateLimit(int $user_id): void
     {
         $max_comments = 30;
         $duration_minutes = 60;
@@ -284,10 +284,10 @@ class TalkCommentMapper extends ApiMapper
             return;
         }
 
-        $rate_limit_time = (new \DateTimeImmutable())
+        $rate_limit_time = (int) (new \DateTimeImmutable())
             ->sub(new \DateInterval(sprintf('PT%dM', $duration_minutes)))
             ->format('U');
-        $oldest_comment = (new \DateTimeImmutable($row['created_at']))->format('U');
+        $oldest_comment = (int) (new \DateTimeImmutable($row['created_at']))->format('U');
 
         $wait = $oldest_comment - $rate_limit_time;
         if ($wait > 0) {
--- a/src/Model/TalkCommentMapper.php
+++ b/src/Model/TalkCommentMapper.php
@@ -3,6 +3,7 @@
 namespace Joindin\Api\Model;
 
 use Exception;
+use Joindin\Api\Exception\RateLimitExceededException;
 use PDO;
 
 class TalkCommentMapper extends ApiMapper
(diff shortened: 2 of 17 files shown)


src/Model/ApiMapper.php
```diff
@@ -78,7 +78,8 @@ public function transformResults(array $results, $verbose)
                     } else {
                         $tz = new DateTimeZone('UTC');
                     }
-                    $entry[$key] = (new DateTime('@' . $row[$value]))->setTimezone($tz)->format('c');
+                    $time = is_numeric($row[$value]) ? '@' . $row[$value] : $row[$value];
+                    $entry[$key] = (new DateTime($time))->setTimezone($tz)->format('c');
```
heiglandreas: Can we please change this to a `DateTimeImmutable` to avoid problems (or to realize them)

src/Model/TalkCommentMapper.php
```diff
@@ -235,6 +236,57 @@ protected function getBasicSQL($include_hidden = false)
...
+    /**
+     * @param int $user_id
+     * @throws RateLimitExceededException
+     */
+    public function checkRateLimit(int $user_id)
+    {
+        if (isset($this->_request)) {
+            $max_comments = (int) $this->_request->getConfigValue('talk_comments_rate_limit_limit', 30);
+            $duration_minutes = (int) $this->_request->getConfigValue('talk_comments_rate_limit_reset', 60);
+        } else {
```
heiglandreas: I'd rather see this else removed by initializing the variable with the default values from the `else` block and only in case there is a request object set the variables default value is modified.

src/Model/TalkCommentMapper.php
```diff
@@ -235,6 +236,57 @@ protected function getBasicSQL($include_hidden = false)
...
+            $max_comments = 30;
+            $duration_minutes = 60;
+        }
+
+        $sql = <<<SQL
+        SELECT MIN(created_at) created_at, COUNT(ID) cnt
```
heiglandreas: Wouldn't this return the same result but with only one select?

```sql
SELECT min(created_at) as created_at, count(id) as cnt
FROM talk_comments
WHERE user_id = :user_id
AND created_at > :min_created_at
GROUP BY user_id
```
execute this using the `user_id` as well as `(new DateTimeImmutable())->sub(new \DateInterval(sprintf('PT%dM', $duration_minutes)))'  as `min_created_at`

src/Model/TalkCommentMapper.php
```diff
@@ -259,10 +311,8 @@ public function save(array $data)
             throw new Exception("Duplicate comment");
         }
 
-        $sql = 'insert into talk_comments (talk_id, rating, comment, user_id, '
-               . 'source, date_made, private, active) '
-               . 'values (:talk_id, :rating, :comment, :user_id, :source, UNIX_TIMESTAMP(), '
-               . ':private, 1)';
+        $sql = 'insert into talk_comments (talk_id, rating, comment, user_id, source, private, active) '
```
heiglandreas: I would rather see the `created_at` being set here explicitly to not depend on the correct setting of the database-server. If we set it explicitly ourselves we know that the timezone-settings are correct (or can't blame the database-server if they aren't).

So the value should be `(new DateTimeImmutable())->setTimezone(new DateTimezone('UTC')->format('Y-m-d\TH:i:s')`

src/Model/TalkCommentMapper.php
```diff
@@ -453,6 +503,7 @@ public function getRawComment($comment_id)
         if ($response) {
             $results = $stmt->fetchAll(PDO::FETCH_ASSOC);
             if (isset($results[0])) {
+                $results[0]['created_at'] = new \DateTimeImmutable($results[0]['created_at']);
```
heiglandreas: This should be `new DateTimeImmutable($results[0]['created_at'], new DateTimeZone('UTC'))`

tools/dbgen/generator.class.php
```diff
@@ -381,11 +381,11 @@ protected function _generateTalkComments($count)
             }
 
             printf(
-                "(%d, %d, '%s', %d, %d, %d, %d, %d, NULL, '%s')",
+                "(%d, %d, '%s', '%s', %d, %d, %d, %d, NULL, '%s')",
                 $talk->id,
                 $rating,
                 $comment,
-                $date_made,
+                (new DateTime('@' . $date_made))->format('Y-m-d H:i:s'),
```
heiglandreas: Please use `DateTimeImmutable` here

svpernova09:

PR #707, opened 2019-10-04
Add an id field to the returned talk data

### LINKED ISSUE #645: Resolving long and short URL formats into Talk ID
The following URLs both point to the event represented by Talk ID `25911`. Is there a way via the API to resolve the appropriate talk ID from the URL?

 * https://joind.in/event/midwest-php-2019/learning-the-php-fig-http-stack
 * https://joind.in/talk/bf577

If not, would it be possible to add the Talk ID as a `<link>` or a response header so that it can be discovered that way? I'd be happy to work on such a thing if that was something that would be interesting.

arjank:
When web requests event talks data, the api response now includes the
talk id (which was already available as part of url fields anyway,
this property just makes sure it is easier to access).

Closes #645

### DIFF
--- a/src/Model/TalkModel.php
+++ b/src/Model/TalkModel.php
@@ -19,6 +19,7 @@ class TalkModel extends BaseModel
     public function getDefaultFields()
     {
         return [
+            'id'                      => 'ID',
             'talk_title'              => 'talk_title',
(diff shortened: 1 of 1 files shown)


iansltx:
Deploy to production with this going out nowish.

PR #710, opened 2019-10-15
add talkmedia when fetching by speaker

bghosts:
Fixes web2 issue #826 missing slide icon

### DIFF
--- a/src/Model/TalkMapper.php
+++ b/src/Model/TalkMapper.php
@@ -313,10 +313,11 @@ class TalkMapper extends ApiMapper
      * @param int $user_id
      * @param int $resultsperpage
      * @param int $start
+     * @param bool $verbose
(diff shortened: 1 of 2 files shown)


src/Model/TalkMapper.php
```diff
@@ -313,10 +313,11 @@ protected function getTracks($talk_id)
      * @param int $user_id
      * @param int $resultsperpage
      * @param int $start
+     * @param bool $verbose
      *
      * @return false|TalkModelCollection
      */
-    public function getTalksBySpeaker($user_id, $resultsperpage, $start)
+    public function getTalksBySpeaker($user_id, $resultsperpage, $start, $verbose = false)
```
localheinz: Why aren’t we using type declarations here?

PR #712, opened 2019-10-15
Fix: Add type declaration

localheinz:
This PR

* [x] adds a type declaration

Follows #710.


localheinz:
Thank you, @BackEndTea and @svpernova09!

PR #714, opened 2019-10-15
Enhancement: Switch to using GitHub actions

localheinz:
This PR

* [x] switches from Travis CI to using GitHub actions
* [x] fixes issues reported via static analysis


localheinz:
Ha!

https://github.com/joindin/joindin-api/pull/714/checks?check_run_id=261434086#step:3:9

>Problem 1
>  - The requested PHP extension ext-intl * is missing from your system. Install or enable PHP's intl

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,31 @@
...
+      matrix:
+        php-binary:
+          - php7.3
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@master
+
+      - name: "Install dependencies with composer"
+        run: $(which composer) install
```
svpernova09: @localheinz would you try throwing a `--ignore-platform-reqs` here to see if this gets us past this issue of misisng ext/intl?

localheinz: Not sure about that, to be honest.

Let's see what we need `ext/intl` for!

localheinz: Running

```
$ composer remove ext-intl
```

followed by

```
$ composer require-checker
```

in #727 yields

```
> composer-require-checker --config-file=composer-require-checker.json
ComposerRequireChecker 2.1.20-250-g14887fc-dev
The following unknown symbols were found:
+----------------+--------------------+
| unknown symbol | guessed dependency |
+----------------+--------------------+
| Transliterator | ext-intl           |
+----------------+--------------------+
Script composer-require-checker --config-file=composer-require-checker.json handling the require-checker event returned with error code 1
```

so we actually need `ext/intl`.

Personally, I'd prefer not to run `composer` with the `--ignore-platform-reqs` options in CI.

Perhaps we can bug @nat with updating the images?

localheinz: Also see https://github.community/t5/GitHub-Actions/Have-PHP-Pecl-installed-by-default/m-p/33371#M1480.

localheinz: Also see https://github.com/actions/virtual-environments/issues/3.

svpernova09:
Got the extensions installed, but still failing that it can't find the exention.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,144 @@
...
+        run: vendor/bin/phpunit --dump-xdebug-filter=build/xdebug-filter.php
+
+      - name: "Run tests with phpunit"
+        run: vendor/bin/phpunit --coverage-clover=build/logs/clover.xml --coverage-text --prepend=build/xdebug-filter.php
+
+      - name: "Download code coverage uploader for Codecov.io"
+        run: curl -s https://codecov.io/bash -o codecov
+
+      - name: "Send code coverage report to Codecov.io"
+        run: bash codecov -t ${{ secrets.CODECOV_TOKEN }}
```
localheinz: The `CODECOV_TOKEN`, taken from https://codecov.io/gh/joindin/joindin-api/settings, needs to be provided as a secret in the repository settings (see https://help.github.com/en/github/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#creating-encrypted-secrets).

Unfortunately, this has the limitation that the secrets aren't available for pull request builds (unless the author has write access).

svpernova09: ` CODECOV_TOKEN` has been added

localheinz:
@svpernova09 

Other than #784, is there anything holding up merging this?

svpernova09:
> Other than #784, is there anything holding up merging this?

Been thinking back and forth on handling code style. As stated in the issue I'm in favor of Style-CI because it takes the burden off me-the-contributor-making-the-pr, and off me-the-code-reviewer reviewing the PR, and off the entire organization as something we can set and forget. 

The Leadership Team had a discussion about this when Style-CI was proposed and the conclusion was to use Style-CI but not remove the tooling to run similar/hopefully the same style changes locally.

I also wouldn't want a code style issue to fail a CI/CD build. Yes the project has adopted a code style but it shouldn't be a sticking point to accept a PR and it shouldn't be a code review item because it's too silly not to automate. Especially when tooling makes it so easy.

I think what I'd prefer is to remove the Github action for "Coding Standards" and continue to allow Style-CI to automatically resolve CS issues.

iansltx:
Been meaning to take a look to see whether we can match StyleCI to PSR-12...and maybe not enforce any other opinions. Just haven't gotten around to it, but once that config is available it should be useful for other projects.

localheinz:
@svpernova09 

This should be good to go then.

The requirement for Travis CI needs to be removed from https://github.com/joindin/joindin-api/settings/branches for `master`, then.

localheinz:
Thank you, @svpernova09!

PR #716, opened 2019-10-15
Enhancement: Use Guzzle client to verify comment

localheinz:
This PR

* [x] uses a `GuzzleHttp\Client` to verify a comment with Akismet

💁‍♂ Since `guzzlehttp/guzzle` is already a dependency, why not use it instead of `curl` directly?

For reference, see https://akismet.com/development/api/#comment-check.

### DIFF
--- a/src/Service/SpamCheckService.php
+++ b/src/Service/SpamCheckService.php
@@ -2,28 +2,37 @@
 
 namespace Joindin\Api\Service;
 
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception;
+
 /**
  * A class that lets you check against an external service (Akismet)
  * for spam in your content
  */
 class SpamCheckService implements SpamCheckServiceInterface
 {
+    private $httpClient;
+
     protected $akismetUrl;
 
     protected $blog;
 
     /**
+     * @param ClientInterface $httpClient
      * @param string $apiKey
      * @param string $blog
      */
-    public function __construct($apiKey, $blog)
+    public function __construct(ClientInterface $httpClient, $apiKey, $blog)
     {
+        $this->httpClient = $httpClient;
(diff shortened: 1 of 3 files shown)


src/Service/SpamCheckService.php
```diff
@@ -33,36 +42,27 @@ public function __construct($apiKey, $blog)
...
+            $response = $this->httpClient->request(
+                'POST',
+                sprintf(
+                    '%s/1.1/comment-check',
+                    $this->akismetUrl
+                ),
+                [
+                    'body' => http_build_query([
+                        'blog' => $this->blog,
+                        'comment_content' => $this->getField('comment', $data),
```
localheinz: Question is: should we even bother sending a request when we already know that the comment is missing from `$data`?

svpernova09: I would say no, we should skip sending the request.

src/Service/SpamCheckService.php
```diff
@@ -33,39 +42,41 @@ public function __construct($apiKey, $blog)
      */
     public function isCommentAcceptable(array $data, $userIp, $userAgent)
     {
-        if (!array_key_exists('comment', $data) || !is_string($data['comment']) || '' === trim($data['comment'])) {
+        try {
```
svpernova09: @localheinz Can you double-check my work here in the conflict resolution is correct?

localheinz:
@svpernova09 

Rebased!

localheinz:
@svpernova09 

Fixed the build!

localheinz:
@svpernova09 

This is ready for another review.

localheinz:
@svpernova09 

Any chance of this getting merged soon?

It has been sitting around for 28 days.

localheinz:
Thank you, @svpernova09!

PR #718, opened 2019-10-16
Fix: Remove PSR-4 mapping for non-existent code

localheinz:
This PR

* [x] removes a PSR-4 mapping for non-existent code


localheinz:
Thank you, @svpernova09!

PR #720, opened 2019-10-16
Fix: Do not return anything from constructor

localheinz:
This PR

* [x] stops attempting to return a value from a constructor


localheinz:
Thank you, @iansltx!

PR #725, opened 2019-10-19

localheinz:
This PR

* [x] requires [`teapot/status-code`](https://github.com/teapot-php/status-code)
* [x] uses a more appropriate HTTP status code


src/Router/VersionedRouter.php
```diff
@@ -71,8 +72,8 @@ function ($val) {
         }
 
         if ($badMethod) {
-            throw new Exception('Method not supported', 415);
```
localheinz: This should probably be `405` instead, shouldn't it?

svpernova09: Agreed

PR #726, opened 2019-10-19

localheinz:
This PR

* [x] requires [`teapot/status-code`](https://github.com/teapot-php/status-code)
* [x] uses a more appropriate HTTP status code

💁‍♂ Replaces #725, which I accidentally closed (without an ability to reopen).


localheinz:
Thank you, @svpernova09!

PR #727, opened 2019-10-19
Enhancement: Use maglnet/composer-require-checker to check for missing dependencies

localheinz:
This PR

* [x] requires [`maglnet/composer-require-checker`](https://github.com/maglnet/ComposerRequireChecker)
* [x] runs `composer-require-checker` as part of `composer check`
* [x] runs `composer-require-checker` on GitHub actions
* [x] adds a basic configuration for `composer-require-checker`
* [x] whitelists dependencies which cannot be loaded with `composer` because they use custom autoloader
* [x] whitelists soft dependencies
* [x] declares missing dependencies in `composer.json`


svpernova09:
I don't understand why this is important. The Readme here: https://github.com/maglnet/ComposerRequireChecker does not do a good job explaining why you'd want this package. Looking at the diff we're adding more extensions for 1 package (for development!).

composer.json
```diff
@@ -12,19 +12,25 @@
...
+    "ext-gd": "*",
     "ext-intl": "*",
     "ext-json": "*",
+    "ext-openssl": "*",
     "ext-pdo": "*",
     "guzzlehttp/guzzle": "^6.3",
     "guzzlehttp/oauth-subscriber": "^0.3.0",
     "michelf/php-markdown": "^1.8",
     "pimple/pimple": "^3.2",
+    "psr/container": "^1.0.0",
```
localheinz: These dependencies are added because they are required - not by `maglnet/composer-require-checker`, but by this application.

svpernova09: So this package ensures anything we're using in our code that isn't explicitly stated in our composer.json, gets added there, and a situation where that is currently happening is we're using `psr/container` but not from our own composer.json? From another dependency that also uses it?

svpernova09:
@localheinz would you please verify this conflict resolution: https://github.com/joindin/joindin-api/pull/727/commits/117546d49bd5f91dad8fe613b7b3337b7992c508

localheinz:
@svpernova09 

Any chance of this getting merged soon?

It has been sitting around for 25 days.

localheinz:
Thank you, @svpernova09!

PR #729, opened 2019-10-19
Fix: Remove empty file

localheinz:
This PR

* [x] removes an empty file which probably has no purpose


svpernova09:
Thanks!

localheinz:
Thank you, @svpernova09!

PR #730, opened 2019-10-19
Fix: Remove duplicated tests

localheinz:
This PR

* [x] removes a duplicated test

💁‍♂ Apart from imports, this tests is exactly the same as [`tests/ContainerFactoryTest.php`](https://github.com/joindin/joindin-api/blob/00a531e89b80dd2797766deed29526326a395ffb/tests/ContainerFactoryTest.php).


localheinz:
Thank you, @svpernova09!

PR #731, opened 2019-10-20
PHPStan passing at level 1

dstockto:
These changes get us to passing PHPStan level 1 - mostly. There's a couple files in scripts, and one function defined in public/index.php that PHPStan calls out for autoloading.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,8 @@
+parameters:
+    excludes_analyse:
+        - */public/index.php
+        - */scripts/*
+    paths:
+        - src
+
+    level: 1
--- a/src/Controller/EventsController.php
+++ b/src/Controller/EventsController.php
@@ -14,10 +14,10 @@ use Joindin\Api\Model\TalkCommentMapper;
 use Joindin\Api\Model\TalkMapper;
 use Joindin\Api\Model\TrackMapper;
 use Joindin\Api\Model\UserMapper;
+use Joindin\Api\Request;
 use Joindin\Api\Service\EventApprovedEmailService;
 use Joindin\Api\Service\EventSubmissionEmailService;
 use PDO;
-use Joindin\Api\Request;
 use Teapot\StatusCode\Http;
 
 class EventsController extends BaseApiController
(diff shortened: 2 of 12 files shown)


localheinz:
@dstockto 

What do you think about splitting this pull request into two commits:

- one that adds/adjusts the primary configuration
- one or more that fix the issues

?

phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
+    excludes_analyse:
+        - */public/index.php
+        - */scripts/*
+    paths:
```
localheinz: What do you think about keeping parameters sorted by name here?

This way it will be more obvious where to add new parameters.

src/Controller/EventsController.php
```diff
@@ -14,10 +14,10 @@
 use Joindin\Api\Model\TalkMapper;
 use Joindin\Api\Model\TrackMapper;
 use Joindin\Api\Model\UserMapper;
+use Joindin\Api\Request;
 use Joindin\Api\Service\EventApprovedEmailService;
 use Joindin\Api\Service\EventSubmissionEmailService;
 use PDO;
-use Joindin\Api\Request;
```
localheinz: Is this change related to increasing the `phpstan` level?

dstockto: It is - one of the things it calls out is unused imports.

phpstan.neon
dstockto: Sure, I can do that.

dstockto:
> @dstockto
> 
> What do you think about splitting this pull request into two commits:
> 
> * one that adds/adjusts the primary configuration
> * one or more that fix the issues
> 
> ?

I can split this up into two.

dstockto:
Configuration PR is here - https://github.com/joindin/joindin-api/pull/735

dstockto:
#736 contains the code changes from this PR. Closing this one.

src/Controller/EventsController.php
dstockto: Hmm. Now I'm not so sure. I was running PHPStan, Phan and Psalm. I'm going to add it back.

PR #732, opened 2019-10-21
Fix: Pass comment directly to SpamCheckService::isCommentAcceptable()

localheinz:
This PR

* [x] simplifies `SpamCheckService::isCommentAcceptable()`
* [x] passes the comment directly to `SpamCheckService::isCommentAcceptable()`

Blocks #716, because fewer tests need to be written when we can avoid all of these unnecessary checks whether the `$data` argument

- is an array
- contains the `comment` key
- contains a `string` value for the `comment` key
- the value for the `comment` key is a non-blank `string`

Since the only value if interest is the value for the `comment` key, we can pass in the comment directly and avoid all of these checks here.

💁‍♂ Probably best reviewed commit by commit.

### DIFF
--- a/src/Service/SpamCheckService.php
+++ b/src/Service/SpamCheckService.php
@@ -25,15 +25,15 @@ class SpamCheckService implements SpamCheckServiceInterface
     /**
      * Check your comment against the spam check service
      *
-     * @param array  $data
+     * @param string $comment
      * @param string $userIp
      * @param string $userAgent
      *
      * @return bool true if the comment is okay, false if it got rated as spam
      */
-    public function isCommentAcceptable(array $data, $userIp, $userAgent)
+    public function isCommentAcceptable(string $comment, $userIp, $userAgent)
     {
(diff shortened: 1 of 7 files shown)


src/Service/SpamCheckService.php
```diff
@@ -25,35 +25,28 @@ public function __construct($apiKey, $blog)
...
      */
-    public function isCommentAcceptable(array $data, $userIp, $userAgent)
+    public function isCommentAcceptable(string $comment, $userIp, $userAgent)
     {
-        if (!array_key_exists('comment', $data) || !is_string($data['comment']) || '' === trim($data['comment'])) {
+        if ('' === trim($comment)) {
             return false;
         }
 
-        $comment = [];
```
localheinz: Needed to inline this variable so I can use `$comment` as parameter.

localheinz:
@svpernova09 

This blocks #716, which has been sitting around for quite a while.

Can we merge this in?

localheinz:
Thank you, @svpernova09!

PR #734, opened 2019-10-23
Don't require track description on track create or edit

iansltx:
If editing a track, providing null/leaving off track description will cause the track description to be kept as-is rather than being edited. Setting to an empty string will clear the track description (which is now a legal action).


iansltx:
@svpernova09 Sorry for the force-push/review dismissal. Updated commit message to reference the issue this is closing as I forgot to add that at first.

iansltx:
Ref: joindin/web#870

iansltx:
ref: joindin/joindin-web2#870

PR #738, opened 2019-10-24
Fix: Actually run phpstan in build

localheinz:
This PR

* [x] adds a `stan` script to `composer.json` and runs it as part of the `check` script
* [x] requires `phpstan/phpstan-shim` as a development dependency
* [x] adds `@property` annotations for magic properties about which `phpstan` complained


localheinz:
@svpernova09 

Any reason this doesn't get merged in?

I think this is rather critical.

svpernova09:
Looks like we need to fix a few more things.

localheinz:
Thank you, @svpernova09!

PR #739, opened 2019-10-24
Fix: Require phpstan/phpstan-shim

localheinz:
This PR

* [x] requires `phpstan/phpstan-shim` as a development dependency

🤷‍♂ Yes, it's somehow installed because it's in `composer.lock`, but it's not listed in `composer.json`. Apparently something went wrong here a while ago.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
   },
   "require-dev": {
     "jakub-onderka/php-parallel-lint": "^1.0",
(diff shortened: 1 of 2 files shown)


localheinz:
@svpernova09 

Any reason this doesn't get merged in?

I think this is rather critical.

localheinz:
Closing in favour of #738.

PR #740, opened 2019-10-24
Fix: Remove unused dependency

localheinz:
This PR

* [x] removes an apparently unused dependency


localheinz:
Thank you, @svpernova09!

PR #741, opened 2019-10-25
Some general code style improvements

paulinevos:
Adding camel case and removing a bunch of unused variables. This is just the result of working on an actual issue, but that branch was becoming quite big so sumbitting this separately (also it's october lol)

### DIFF
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -15,13 +15,13 @@ use Teapot\StatusCode\Http;
 
 class UsersController extends BaseApiController
 {
-    protected $user_mapper;
+    protected $userMapper;
 
-    private $user_registration_email_service;
+    private $userRegistrationEmailService;
 
     public function getAction(Request $request, PDO $db)
     {
-        $user_id = $this->getItemId($request);
+        $userId = $this->getItemId($request);
 
         // verbosity
         $verbose = $this->getVerbosity($request);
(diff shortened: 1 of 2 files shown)


src/Controller/UsersController.php
```diff
@@ -466,25 +464,6 @@ public function setTrusted(Request $request, PDO $db)
         $view->setResponseCode(Http::NO_CONTENT);
     }
 
-    /**
```
paulinevos: Not sure why these changes are in here, will clean (will be part of other MR)

.gitignore
```diff
@@ -11,3 +11,7 @@ vendor/.git
 node_modules
 vendor
 diff.txt
+.phpcs-cache
+.phpunit.result.cache
+tests/.phpunit.results.cache
```
localheinz: This should not be necessary, see:

https://github.com/joindin/joindin-api/blob/f7dede99cbe0c325565d64626999c4797ef4d68f/phpunit.xml.dist#L11:

.gitignore
```diff
@@ -11,3 +11,7 @@ vendor/.git
 node_modules
 vendor
 diff.txt
+.phpcs-cache
```
localheinz: This should not be necessary, see:

https://github.com/joindin/joindin-api/blob/f7dede99cbe0c325565d64626999c4797ef4d68f/composer.json#L67

.gitignore
paulinevos: Ah, I think I still had an unstaged cache file from months back before that was implemented :)

Will drop the commit.

PR #746, opened 2019-10-25
Fix user ID bug

paulinevos:
There was a mix-up between the logged in user ID and the user ID that was meant to be deleted. This resulted in an exception, because we were checking if the user to be deleted was an admin user instead of the actual logged in user.


paulinevos:
Done did!

PR #751, opened 2019-10-26
Fix: Remove unnecessary arguments and options

localheinz:
This PR

* [x] removes unnecessary arguments and options when invoking `phpunit` from the corresponding `composer` script


localheinz:
Thank you, @svpernova09!

PR #752, opened 2019-10-26
Fix: Actually collect coverage in coverage script

localheinz:
This PR

* [x] actually collects useful coverage information in the `coverage` script
* [x] runs the `coverage` script instead of the `test` script in the `check` script
* [x] sends code coverage information to https://codecov.io

💁‍♂ Not sure whether you have a preference for a service that collects code coverage information, but codecov is rather simple to set up and has nice coverage reports (with diffs) in pull requests, when enabled. 

Please take a look at 

* https://github.com/marketplace/codecov
* https://codecov.io/gh/joindin/joindin-api

as this needs setup from the maintainers' side.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,8 +56,8 @@
       "@test"
     ],
     "coverage": [
-      "git diff origin/master... -- > diff.txt",
-      "diffFilter --phpunit diff.txt build/logs/clover.xml 80"
+      "phpunit --dump-xdebug-filter=build/xdebug-filter.php",
+      "phpunit --coverage-clover=build/logs/clover.xml --coverage-text --prepend=build/xdebug-filter.php"
     ],
     "fix": "phpcbf -p .",
     "lint": "parallel-lint --exclude vendor .",
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
       "@lint",
       "@sniff",
       "@security",
-      "@test"
+      "@coverage"
     ],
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -53,11 +53,11 @@
       "@lint",
       "@sniff",
       "@security",
-      "@test"
+      "@coverage"
     ],
     "coverage": [
-      "git diff origin/master... -- > diff.txt",
-      "diffFilter --phpunit diff.txt build/logs/clover.xml 80"
+      "phpunit --dump-xdebug-filter=build/xdebug-filter.php",
```
localheinz: For reference, see https://phpunit.readthedocs.io/en/8.4/code-coverage-analysis.html#speeding-up-code-coverage-with-xdebug.

composer.json
```diff
@@ -53,11 +53,11 @@
       "@lint",
       "@sniff",
       "@security",
-      "@test"
+      "@coverage"
     ],
     "coverage": [
-      "git diff origin/master... -- > diff.txt",
-      "diffFilter --phpunit diff.txt build/logs/clover.xml 80"
```
localheinz: I had removed this with #740, wasn't aware this was used here.

However, I believe that a service that collects the code coverage changes over time and displays the diff in the pull request is more useful.

What do you think?

localheinz:
Thank you, @svpernova09!

Issue #618, opened 2019-02-03
Bump SwiftMailer to v6


iansltx:
We'll need to hold off on merging this 'til we're targeting PHP 7 in prod, but the requisite dev work can happen now.

Looks like SwiftMailer is used in one place (BaseEmailService) and none of the classes that inherit from the base use SwiftMailer classes directly so this shouldn't be a huge leap. Going to self-assign this, but if someone beats me to it I'm not gonna complain :)

PR #753, opened 2019-10-27
Upgrade Swiftmailer

svpernova09:
Resolves #698 #618


svpernova09:
Also verified via local dev configured to use Mail hog that emails are still being sent by sending user name and password reset emails from web2 frontend.

PR #755, opened 2019-10-27
Add ability to provide a reason for event rejection & send email

svpernova09:
Followed the same format we currently do to send an event approval email but modified it to send email when an event is rejected w/ a reason.

# Questions to be Answered
* Should we pass the `reason` to the DELETE endpoint in a body field as opposed to how I'm currently doing it with a parameter?


src/Controller/EventsController.php
```diff
@@ -733,16 +734,25 @@ public function rejectAction(Request $request, PDO $db)
...
         }
 
-        $result = $event_mapper->reject($event_id, $request->user_id);
+        $result = $event_mapper->reject($event_id, $request->user_id, $reason);
         if (!$result) {
             throw new Exception("This event cannot be rejected", Http::BAD_REQUEST);
         }
 
+        // Only send an email if we provided a reason
+        if (!empty($reason)) {
```
ericpoe: [nb] Why check for non-empty? Wouldn't `if ($reason) {` suffice?

localheinz:
@svpernova09 

What do you think about adding tests?

src/Controller/EventsController.php
localheinz: Can we avoid using `empty()` here and instead use a stricter verification?

src/Controller/EventsController.php
```diff
@@ -733,16 +734,25 @@ public function rejectAction(Request $request, PDO $db)
...
+            $event        = $event_mapper->getEventById($event_id, true, false)['events'][0];
+            $recipients   = array_merge([$this->config['email']['from']], $event_mapper->getHostsEmailAddresses($event_id));
+            $emailService = new EventRejectedEmailService($this->config, $recipients, $event);
```
localheinz: To me this looks like an instance of `EventRejectedEmailService` is a hard dependency of this controller action. 

Instead of creating it here, can we inject it via constructor injection?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Joindin\Api\Service;
+
+use DateTime;
+
+class EventRejectedEmailService extends BaseEmailService
```
localheinz: This class doesn't have any tests, can we add them?

localheinz: Can we mark this class as `final`, or are there any reasons to extend from it?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    protected $event;
```
localheinz: Do we need `protected` visibility here, or can these fields be `private`?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+    protected $website_url;
```
localheinz: Can we avoid using `snake_case` here?

See #741.

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function __construct(array $config, array $recipients, array $event)
+    {
+        // set up the common stuff first
```
localheinz: Can we remove this comment?

It does not add any value.

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+        parent::__construct($config, $recipients);
+
+        $this->event       = $event;
+        $this->website_url = $config['website_url'];
+    }
+
+    public function sendEmail()
```
localheinz: Can we add a `void` return type declaration here?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $this->setSubject('Event Rejected');
+
+        $date         = new DateTime($this->event['start_date']);
+        $replacements = [
```
localheinz: Can we inline this variable?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+            "title"        => $this->event['name'],
```
localheinz: Do we need to use double quotes here, or will single quotes suffice?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+            "description"  => $this->event['description'],
+            "reason"       => $this->event['rejection_reason'],
+            "date"         => $date->format('jS M, Y'),
+            "contact_name" => $this->event['contact_name'],
+            "website_url"  => $this->website_url,
+            "event_url"    => $this->website_url . '/event/' . $this->event['url_friendly_name'] . '/edit',
+        ];
+
+        $messageBody = $this->parseEmail("eventRejected.md", $replacements);
```
localheinz: Do we need to use double quotes here, or will single quotes suffice?

src/View/emails/eventRejected.md
```diff
@@ -0,0 +1,7 @@
+Your event, *[title]* has been rejected due to the following reason(s):
+
+[reason]
+
+If we have requested changes to your event you can make them [here]([event_url]).
+
+If you have any questions please reply to this email.
```
localheinz:
```suggestion
If you have any questions, please reply to this email.
```

src/View/emails/eventRejected.md
```diff
@@ -0,0 +1,7 @@
+Your event, *[title]* has been rejected due to the following reason(s):
```
localheinz:
```suggestion
Your event with the "[title]" has been rejected due to the following reason(s):
```

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+        $messageHTML = $this->markdownToHtml($messageBody);
```
localheinz: Can we rename this variable to `$htmlBody`?

src/Service/EventRejectedEmailService.php
localheinz: Can we rename this variable to `$body`?

src/Controller/EventsController.php
```diff
@@ -733,16 +734,25 @@ public function rejectAction(Request $request, PDO $db)
...
         $event_id     = $this->getItemId($request);
         $event_mapper = new EventMapper($db, $request);
+        $reason = filter_var($request->getParameter('reason'), FILTER_SANITIZE_STRING);
 
         if (!$event_mapper->thisUserCanApproveEvents()) {
             throw new Exception("You are not allowed to reject this event", Http::FORBIDDEN);
         }
 
-        $result = $event_mapper->reject($event_id, $request->user_id);
+        $result = $event_mapper->reject($event_id, $request->user_id, $reason);
```
localheinz: What do you think about wrapping the arguments to this method call?

src/Controller/EventsController.php
```diff
@@ -733,16 +734,25 @@ public function rejectAction(Request $request, PDO $db)
...
         if (!$result) {
             throw new Exception("This event cannot be rejected", Http::BAD_REQUEST);
         }
 
+        // Only send an email if we provided a reason
+        if (!empty($reason)) {
+            $event        = $event_mapper->getEventById($event_id, true, false)['events'][0];
```
localheinz: Can we even be sure that we have an event for `$event_id` here, that is, blindly assume that the return value of `getEventById()` will be an array that has an `events` key, where the corresponding value is an array that that has an entry for key `0`?

src/Service/EventRejectedEmailService.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Joindin\Api\Service;
+
+use DateTime;
+
+class EventRejectedEmailService extends BaseEmailService
+{
+    protected $event;
+    protected $website_url;
+
+    public function __construct(array $config, array $recipients, array $event)
```
localheinz: Why would we want to inject 

- a list of recipients
- event information

via constructor?

These arguments should probably be passed to a method that will then actually construct and send a message.

svpernova09:
> What do you think about adding tests?

I didn't see any tests for the counter version of this so I didn't take the time to sort out testing. I wasn't sure how to go about writing tests as the past few times I've tried I get really frustrated and throw everything away. We will happily take tests in iteration but would like to provide this feature to admins sooner than later.

src/Service/EventRejectedEmailService.php
svpernova09: I'm not interested in solving the implementation issues of this class in this PR, I'm primarily focused on shipping a feature. I would welcome any code cleanup outside of this scope.

src/Controller/EventsController.php
svpernova09: I agree with your concerns but we've not run into issues that I'm aware of. (No one has reported issues rejecting events, and there have certainly been a few rejected over the past few weeks.)

src/Controller/EventsController.php
svpernova09: I typically will do this once the line hits 120 chars

src/Service/EventRejectedEmailService.php
svpernova09: I would prefer to keep this file similar to the Event Accept version until we want to refactor them both.

svpernova09: There's a *ton* of places we could be using single instead of double. I'd rather tackle that in a separate PR.

src/Controller/EventsController.php
svpernova09: I get it, but we're doing the exact same thing with `EventApprovedEmailService`. I'd rather ship and iterate than get botted down in detail.

src/Controller/EventsController.php
svpernova09: Why wouldn't empty be strict enough? The variable will only ever be a string, an empty string, or false.

src/Controller/EventsController.php
localheinz: It could also be a `string` consisting of whitespace only.

src/Controller/EventsController.php
localheinz: It could also be a `string` consisting of whitespace only.

src/Controller/EventsController.php
localheinz: So we prefer to continue shipping software with the same quality, rather than turning the ship around?

src/Service/EventRejectedEmailService.php
localheinz: That means that we add more work first so someone else can sort it out later?

src/Service/EventRejectedEmailService.php
localheinz: [codecov](https://codecov.io/gh/joindin/joindin-api)

src/Controller/EventsController.php
mcneely: trim() would fix that.

src/Model/EventMapper.php
```diff
@@ -1108,10 +1110,11 @@ public function approve($event_id, $reviewing_user_id)
      *
      * @param  integer $event_id
      * @param  integer $reviewing_user_id The user who rejected the event
+     * @param  string $reason_Details for rejection
      *
      * @return boolean
      */
-    public function reject($event_id, $reviewing_user_id)
+    public function reject($event_id, $reviewing_user_id, $reason)
```
localheinz: What do you think about adding a type declaration here?

```suggestion
    public function reject($event_id, $reviewing_user_id, string $reason)
```

src/Model/EventMapper.php
```diff
@@ -1108,10 +1110,11 @@ public function approve($event_id, $reviewing_user_id)
      *
      * @param  integer $event_id
      * @param  integer $reviewing_user_id The user who rejected the event
+     * @param  string $reason_Details for rejection
```
localheinz:
```suggestion
     * @param  string $reason Details for rejection
```

PR #757, opened 2019-10-27
Fix: Make use of __DIR__ constant

localheinz:
This PR

* [x] makes use of the `__DIR__` constant when referencing file paths

### DIFF
--- a/src/public/index.php
+++ b/src/public/index.php
@@ -9,7 +9,7 @@ use Teapot\StatusCode\Http;
 
 include __DIR__ . '/../../vendor/autoload.php';
 if (!function_exists('apache_request_headers')) {
-    include '../inc/nginx-helper.php';
+    include __DIR__ . '/../inc/nginx-helper.php';
 }
 if (!ini_get('date.timezone')) {
     date_default_timezone_set('UTC');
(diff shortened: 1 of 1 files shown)


src/public/index.php
```diff
@@ -9,7 +9,7 @@
 
 include __DIR__ . '/../../vendor/autoload.php';
 if (!function_exists('apache_request_headers')) {
-    include '../inc/nginx-helper.php';
+    include __DIR__ . '/../inc/nginx-helper.php';
```
localheinz: This should probably be `require`.

src/public/index.php
```diff
@@ -41,7 +41,7 @@ function handle_exception(\Throwable $e)
 // config setup
 define('BASEPATH', '.');
 $config = [];
-include '../config.php';
+include __DIR__. '/../config.php';
```
localheinz: This should probably be `require`.

src/public/index.php
```diff
@@ -51,7 +51,7 @@ function handle_exception(\Throwable $e)
 
 // database setup
 $db = [];
-include '../database.php';
+include __DIR__ . '/../database.php';
```
localheinz: This should probably be `require`.

svpernova09:
Thanks for these changes.

localheinz:
Thank you, @svpernova09!

PR #759, opened 2019-10-27
Enhancement: Move application entry-point out of src directory

localheinz:
This PR

* [x] moves the application entry-point from `src/public` to `public`


localheinz:
Thank you, @svpernova09!

PR #760, opened 2019-10-27
Enhancement: Move include out of src directory

localheinz:
This PR

* [x] moves `src/inc` to `inc`


svpernova09:
I think I'd rather refactor away from using `$headers = array_change_key_case(apache_request_headers(), CASE_LOWER);` to get the headers? Looks like we're using our own request class and I don't see any header stuff ...

localheinz:
Thank you, @svpernova09!

PR #762, opened 2019-10-28
Enhancement: Raise phpstan level from 1 to 2

localheinz:
This PR

* [x] raises the `phpstan` level from `1` to `2`
* [x] fixes issues reported by `phpstan`

❗️ Blocks #767.


svpernova09:
I would expect us to current support PSR-2, and be moving towards PSR-12 https://www.php-fig.org/psr/psr-12/

Issue #650, opened 2019-04-08
Add "delete all comments for user" endpoint (for site admins only)


iansltx:
This would be available as an "in case of emergency break glass" endpoint, in cases of excessive spamming (e.g. what happened this evening). The endpoint should return a payload of all comment information that was deleted as an audit trail, so it needs to grab that comment information before running `DELETE FROM talk_comments WHERE user_id = ?`.

paulinevos:
Are you sure about returning all the comment information? Just wanted to pick this up but if someone's spammed thousands of comments... would that not be a bit of a big payload? Could this audit trail be written somewhere else? And what should be in there?

iansltx:
My line of reasoning here was that, if the purpose is actually to delete comments, there isn't really a better place to put all that info. Would be a huge blob to log, potentially exceeding size limits unless you threw each comment on its own log line. Throwing comments back in the database would defeat the purpose a bit. Only other thing I can think of would be sending an email of some sort to whoever kicked off the call, which could work/would probably be a better audit trail vs. just showing a response. Maybe throw the info into a CSV attachment or some such.

As for what to include, assuming we used the email method (so we're actually removing things from the system of record), I'd say all columns in the table. That would allow for reconstruction of those comments easily if needed later.

iansltx:
(FWIW if I remember correctly from when i added this issue we were looking at on the order of 1k comments, all of them relatively short, so definitely within email attachment range, but definitely rather large for an HTTP response)

paulinevos:
How about... I just delete them for now and the audit trail is an iteration :v

paulinevos:
Also... is this talk comments, event comments, or both?

iansltx:
I'm fine reviewing the deletion bit as its own first step, but to avoid some idiot (me) more easily destroying 13/10 good comments I'd like to have an audit trail pre-merge. I don't mind writing those bits if need be though.

paulinevos:
Sounds good.

iansltx:
Also, saw your other comment in my inbox but not here. Odd. Anyway, from what I saw talks were the only ones getting spammed, but feel free to throw event comments in here too.

paulinevos:
... maybe next time 👀

paulinevos:
Oh btw, probably because I deleted that comment cause I saw it was talk comments in the sql you put.

heiglandreas:
Any news here?

paulinevos:
[...]

PR #763, opened 2019-10-28
Delete all talk comments for user

paulinevos:
Solves #650 (partially, does not include the logging part as discussed with @iansltx).

Introduces a `DELETE` verb on `users/{userId}/talk_comments` in order to delete all a user's talk comments in one go. Only accessible for admins. Also introduces some new exceptions that I noticed are used in a lot of different places. Might open a PR to add usages of those in other places.


src/Controller/TalkCommentsController.php
```diff
@@ -15,6 +15,11 @@
 class TalkCommentsController extends BaseApiController
 // @codingStandardsIgnoreEnd
 {
+    /**
+     * @var TalkCommentMapper
+     */
+    protected $commentMapper;
```
localheinz: Will `private` visibility suffice here?

src/Controller/TalkCommentsController.php
```diff
@@ -189,4 +194,13 @@ public function updateComment(Request $request, PDO $db)
 
         return $commentMapper->getCommentById($commentId);
     }
+
+    private function getCommentMapper(Request $request, PDO $db): TalkCommentMapper
+    {
+        if ($this->commentMapper === null) {
```
localheinz: Unless we assign `$this->commentMapper` a value, this condition will always evaluate to `true`.

src/Controller/TalkCommentsController.php
```diff
@@ -189,4 +194,13 @@ public function updateComment(Request $request, PDO $db)
...
+            return new TalkCommentMapper($db, $request);
```
localheinz: Perhaps you meant this?

```suggestion
            $this->commentMapper = new TalkCommentMapper($db, $request);
```

src/Controller/UsersController.php
```diff
@@ -19,6 +21,11 @@ class UsersController extends BaseApiController
 
     private $userRegistrationEmailService;
 
+    /**
+     * @var TalkCommentMapper
+     */
+    protected $talkCommentMapper;
```
localheinz: Will `private` visibility suffice here?

src/Controller/UsersController.php
```diff
@@ -408,6 +415,30 @@ public function passwordReset(Request $request, PDO $db)
         }
     }
 
+    /**
+     * @throws Exception
+     */
+    public function deleteTalkComments(Request $request, PDO $db): void
+    {
+        if (false == ($request->getUserId())) {
```
localheinz:
```suggestion
        if (null === $request->getUserId()) {
```

src/Controller/UsersController.php
```diff
@@ -476,6 +507,20 @@ public function getUserMapper(PDO $db, Request $request)
         return $this->userMapper;
     }
 
+    public function setTalkCommentMapper(TalkCommentMapper $talkCommentMapper): void
+    {
+        $this->talkCommentMapper = $talkCommentMapper;
+    }
+
+    public function getTalkCommentMapper(PDO $db, Request $request): TalkCommentMapper
```
localheinz: Since the return value of this method is currently never used, perhaps it would make more sense to rename this method to `initializeTalkCommentMapper()`, and add a `void` return type declaration?

src/Exception/AuthenticationException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Joindin\Api\Exception;
+
+use \Exception;
```
localheinz:
```suggestion
use Exception;
```

src/Exception/AuthenticationException.php
```diff
@@ -0,0 +1,18 @@
...
+use \Throwable;
```
localheinz:
```suggestion
use Throwable;
```

src/Exception/AuthenticationException.php
```diff
@@ -0,0 +1,18 @@
...
+
+final class AuthenticationException extends Exception implements Throwable
+{
+    private const MESSAGE = 'You must be logged in to perform this operation.';
+
+    public static function forUnauthenticatedUser(string $message = null): self
+    {
+        return new static($message ?? self::MESSAGE, 401);
```
localheinz: What do you think about using the corresponding constant from `teapot/status-code`?

```diff
diff --git a/src/Exception/AuthenticationException.php b/src/Exception/AuthenticationException.php
index 4becf9f..2635e47 100644
--- a/src/Exception/AuthenticationException.php
+++ b/src/Exception/AuthenticationException.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Joindin\Api\Exception;

 use \Exception;
+use Teapot\StatusCode\Http;
 use \Throwable;

 final class AuthenticationException extends Exception implements Throwable
@@ -13,6 +14,6 @@ final class AuthenticationException extends Exception implements Throwable

     public static function forUnauthenticatedUser(string $message = null): self
     {
-        return new static($message ?? self::MESSAGE, 401);
+        return new static($message ?? self::MESSAGE, Http::UNAUTHORIZED);
     }
 }
```

src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Joindin\Api\Exception;
+
+use \Exception;
```
localheinz:
```suggestion
use Exception;
```

src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,18 @@
...
+use \Throwable;
```
localheinz:
```suggestion
use Throwable;
```

src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,18 @@
...
+
+final class AuthorizationException extends Exception implements Throwable
+{
+    private const MESSAGE = 'This operation requires %s privileges.';
+
+    public static function forNonAdministrator(): self
+    {
+        return new static(sprintf(self::MESSAGE, 'admin'), 403);
```
localheinz: What do you think about using the corresponding constant from `teapot/status-code`?

```diff
diff --git a/src/Exception/AuthorizationException.php b/src/Exception/AuthorizationException.php
index 6e3b6e8..e22ea7d 100644
--- a/src/Exception/AuthorizationException.php
+++ b/src/Exception/AuthorizationException.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Joindin\Api\Exception;

 use \Exception;
+use Teapot\StatusCode\Http;
 use \Throwable;

 final class AuthorizationException extends Exception implements Throwable
@@ -13,6 +14,6 @@ final class AuthorizationException extends Exception implements Throwable

     public static function forNonAdministrator(): self
     {
-        return new static(sprintf(self::MESSAGE, 'admin'), 403);
+        return new static(sprintf(self::MESSAGE, 'admin'), Http::FORBIDDEN);
     }
 }
```

src/Exception/AuthorizationException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Joindin\Api\Exception;
+
+use \Exception;
+use \Throwable;
+
+final class AuthorizationException extends Exception implements Throwable
```
localheinz: The `Throwable` interface is already implemented by `Exception`, can we remove the `implements`?

src/Exception/Comment/TalkCommentDeletionException.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace Joindin\Api\Exception\Comment;
+
+use Exception;
+use Teapot\StatusCode\Http;
+use Throwable;
+
+final class TalkCommentDeletionException extends Exception implements Throwable
```
localheinz: The `Throwable` interface is already implemented by `Exception`, can we remove the `implements`?

src/Exception/AuthenticationException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Joindin\Api\Exception;
+
+use \Exception;
+use \Throwable;
+
+final class AuthenticationException extends Exception implements Throwable
```
localheinz: Perhaps it makes sense to extend a more specific exception, for example,  `RuntimeException`?

src/Exception/AuthorizationException.php
localheinz: Perhaps it makes sense to extend a more specific exception, for example,  `RuntimeException`?

src/Exception/Comment/TalkCommentDeletionException.php
localheinz: Perhaps it makes sense to extend a more specific exception, for example,  `RuntimeException`?

src/config/routes/2.1.php
```diff
@@ -470,6 +470,16 @@
                 'PUT',
             ],
     ],
+
+    [
+        'path'       => '/users/[0-9]+/talk_comments',
+        'controller' => UsersController::class,
+        'action'     => 'deleteTalkComments',
+        'verbs'      =>
+            [
```
localheinz: Do we want to be consistent and continue to use the weird wrapping, or perhaps adjust it?

```diff
diff --git a/src/config/routes/2.1.php b/src/config/routes/2.1.php
index 425e142..c41ac7f 100644
--- a/src/config/routes/2.1.php
+++ b/src/config/routes/2.1.php
@@ -42,10 +42,9 @@ return [
         'path'       => '/events/[\\d]+/hosts/[\\d]+',
         'controller' => EventHostsController::class,
         'action'     => 'removeHostFromEvent',
-        'verbs'      =>
-            [
-                'DELETE',
-            ],
+        'verbs'      => [
+            'DELETE',
+        ],
     ],
     [
         'path'       => '/events(/[^/]+)*/comments/reported',
```

tests/Controller/UsersControllerTest.php
```diff
@@ -149,6 +152,90 @@ public function testDeleteUserWithAdminAccessDeletesSuccessfully()
         $this->assertNull($usersController->deleteUser($request, $db));
     }
 
+    public function testDeleteTalkCommentsWithoutBeingLoggedInThrowsException(): void
+    {
+        $this->expectException(AuthenticationException::class);
+        $this->expectExceptionMessage('You must be logged in to perform this operation.');
+        $this->expectExceptionCode(Http::UNAUTHORIZED);
```
localheinz: What do you think about setting these expectations as late as possible?

For reference, see https://thephp.cc/dates/2019/11/symfonycon/phpunit-best-practices.

src/config/routes/2.1.php
paulinevos: Hm, not sure, do we have a standard for this? Also not sure if this is picked up by the cs fixer but will check.

localheinz: Currently we do not have anything configured for `phpcs`, so please ignore my comment.

tests/Controller/UsersControllerTest.php
paulinevos: Yep, sounds good to me. Might add another commit moving the expectations in the other cases then

src/Controller/TalkCommentsController.php
paulinevos: Should do, not sure why it's `protected`. I never make anything `protected`, lol.

PR #764, opened 2019-10-28
Fix: Inject SpamCheckService into controllers

localheinz:
This PR

* [x] injects implementations of the `SpamCheckServiceInterface` into, rather than creating them in dependent services

❗️Blocks #716.

### DIFF
--- a/src/ContainerFactory.php
+++ b/src/ContainerFactory.php
@@ -85,8 +85,11 @@ class ContainerFactory
                 return new EmailsController($config);
             });
 
-            $container[EventCommentsController::class] = $container->factory(function ($c) use ($config) {
-                return new EventCommentsController($config);
+            $container[EventCommentsController::class] = $container->factory(static function ($container) use ($config): EventCommentsController {
+                return new EventCommentsController(
+                    $container[SpamCheckServiceInterface::class],
+                    $config
+                );
             });
 
             $container[EventHostsController::class] = $container->factory(function ($c) use ($config) {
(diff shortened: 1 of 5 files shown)


src/ContainerFactory.php
```diff
@@ -85,8 +85,11 @@ public static function build(array $config, $rebuild = false): ContainerInterfac
                 return new EmailsController($config);
             });
 
-            $container[EventCommentsController::class] = $container->factory(function ($c) use ($config) {
-                return new EventCommentsController($config);
+            $container[EventCommentsController::class] = $container->factory(static function ($container) use ($config): EventCommentsController {
```
localheinz: These closures can be `static`, we can use `$container` instead of `$c` as a parameter name, and we can add a return type declaration.

tests/Controller/TalksControllerTest.php
```diff
@@ -869,18 +895,24 @@ public function testCommentOnTalk(
 
         $talks_comment_email->method('sendEmail');
 
-        $talks_controller = $this->getMockBuilder(TalksController::class)
-            ->setMethods(['getTalkCommentEmailService'])
-            ->getMock();
+        $talks_controller = new class(new NullSpamCheckService(), $talks_comment_email) extends TalksController {
```
localheinz: Using an anonymous class to override the behaviour of the `getTalkCommentEmailService()` method makes a lot more sense here than trying to create a partial mock.

Of course, this would be a lot better if we injected all dependencies instead of creating them in services themselves.

From the general test setup it's not even clear what the system under test is.

localheinz:
@iansltx @dstockto @svpernova09 @zghosts 

Any chance of this getting merged in soon?

localheinz:
Thank you, @svpernova09!

PR #769, opened 2019-10-29
Fix: Configure path to cache file via phpcs.xml.dist

localheinz:
This PR

* [x] configures the path to the cache file used by `phpcs` via `phpcs.xml.dist`
* [x] ensures that the `build` directory exists in the `fix` and `sniff` scripts


localheinz:
Thank you, @svpernova09!

PR #771, opened 2019-10-29
Fix: Cache directory for composer

localheinz:
This PR

* [x] fixes the cache directories used by `composer` on Travis


localheinz:
Thank you, @svpernova09!

PR #772, opened 2019-10-29
Enhancement: Use PSR-12 instead of PSR-2 coding standard

localheinz:
This PR

* [x] uses the `PSR12` coding standard instead of `PSR2`
* [x] runs `composer fix`

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -4,7 +4,7 @@
         name="JoindIn"
         xsi:noNamespaceSchemaLocation="vendor/squizlabs/php_codesniffer/phpcs.xsd"
 >
- <description>The Joind.in coding standard (PSR2, no namespace).</description>
+ <description>The Joind.in coding standard</description>
 
  <exclude-pattern>**/config.php</exclude-pattern>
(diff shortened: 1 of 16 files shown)


inc/nginx-helper.php
```diff
@@ -1,8 +1,11 @@
 <?php
+
+declare(strict_types=1);
```
localheinz: The `Doctrine` coding standard is probably a bit to strict for this project.

With a lack of tests, I would probably not apply this standard.

localheinz:
After running 

```
$ composer fix
```

running

```
$ composer sniff
```

fails with a ton of errors that would need to get fixed manually.

As an alternative to using `squizlabs/php_codesniffer`, we could use `friendsofphp/php-cs-fixer`.

It has the advantage that it comes with a single command line tool, `phpcs`, and only consists of fixers - that is, it will only complain about code style issues that it is capable of fixing.

svpernova09:
I would expect us to current support PSR-2, and be moving towards PSR-12 https://www.php-fig.org/psr/psr-12/

localheinz:
Thank you, @svpernova09!

PR #775, opened 2019-10-30
Fix: Mark test classes as final

localheinz:
This PR

* [x] marks test classes as `final`

💁‍♂ There's probably no reason to extend from these classes.


localheinz:
Thank you, @svpernova09!

PR #776, opened 2019-10-30
Fix: Mark class as abstract

localheinz:
This PR

* [x] marks what appears to be an abstract test case as `abstract`
* [x] marks `protected` method declared in this `abstract` test case and never overridden by extending classes as `final`


svpernova09:
Thanks1

localheinz:
Thank you, @svpernova09!

PR #777, opened 2019-10-31
Fix: List build steps instead of running 'composer check' on Travis

localheinz:
This PR

* [x] lists the individual build steps instead of running `composer check` on Travis

💁‍♂ Yes, this brings a bit of duplication, but it also makes it more transparent which steps we run, and which step exactly failed.


localheinz:
Thank you, @svpernova09!

PR #778, opened 2019-10-31
Fix: Cache directories

localheinz:
This PR

* [x] fixes the cache directories for Travis CI 

💁‍♂ Apparently something went wrong when merging #771.


localheinz:
Thank you, @svpernova09!

PR #773, opened 2019-10-30
Enhancement: Raise phpstan level from 4 to 5

localheinz:
This PR

* [x] raises the `phpstan` level from `4` to `5`
* [ ] fixes issues reported by `phpstan`

PR #780, opened 2019-11-05
Fix: Drop phpstan level from 5 to 4

localheinz:
This PR

* [x] reverts #773 

💁‍♂ This one wasn't ready yet - as pointed out, it fails when running locally, but it passes when running on Travis (which is rather odd).


localheinz:
Thank you, @svpernova09!

PR #782, opened 2019-11-05
Fix: Remove dist extension

localheinz:
This PR

* [x] removes useless `.dist` extensions from `phpcs` and `phpunit` configuration

💁‍♂ They might be useful when

- the non-`dist` versions are ignored (not the case, see [`.gitignore`](https://github.com/joindin/joindin-api/blob/da24240d325c18c37ff608222f219a7cd7366194/.gitignore)) _and_ 
- someone has a need to modify these files locally

Personally, I have never found a reason to do so.


localheinz:
Thank you, @svpernova09!

PR #783, opened 2019-11-05
Unit tests for Header class

dstockto:
This gets the Header class to 100% code coverage -- but...

The class appears to be a copy, or really close to a copy of the Guzzle 3 Header class. It might be good to go ahead and just use Guzzle's code via composer (although maybe not Guzzle 3).

The final test (coverageForParseParams) is there solely to hit coverage for the continue line for Header.php line 173 in the parseParams method. But it's a garbage test in that it doesn't help me or likely anyone else know or understand how parseParams works or should be used.


svpernova09:
> Let's remove it and use PSR-7 implementation from a 3rd-party library instead!

I'm good with this.

PR #787, opened 2019-11-05
Fix: Remove build.xml

localheinz:
This PR

* [x] removes an apparently unused, but at least outdated `build.xml`

💁‍♂ `phing/phing` is not even a dependency - is this file used?


svpernova09:
@ericpoe is this used for the migrations?

PR #791, opened 2019-11-10
Enhancement: Also run phpstan against tests

localheinz:
This PR

* [x] also runs `phpstan` against tests
* [x] fixes issues reported via static analysis

❗️ Blocks #807.

💁‍♂ Probably best reviewed commit by commit.

PR #792, opened 2019-11-10
Fix: Move mockPDO into a file that can be autoloaded

localheinz:
This PR

* [x] moves a class that was declared in `tests/bootstrap.php` into a file that can be autoloaded

❗️ Blocks #791.


localheinz:
Cherry-picked into #791.

PR #798, opened 2019-11-11
Enhancement: Introduce explaining variable

localheinz:
This PR

* [x] introduces explaining variables instead of invoking `$request->getView()` multiple times in a row

### DIFF
--- a/src/Controller/ApplicationsController.php
+++ b/src/Controller/ApplicationsController.php
@@ -88,8 +88,11 @@ class ApplicationsController extends BaseApiController
         $clientId     = $clientMapper->createClient($app);
 
         $uri = $request->base . '/' . $request->version . '/applications/' . $clientId;
-        $request->getView()->setResponseCode(Http::CREATED);
-        $request->getView()->setHeader('Location', $uri);
+
+        $view = $request->getView();
+
+        $view->setResponseCode(Http::CREATED);
+        $view->setHeader('Location', $uri);
 
         $mapper    = $this->getClientMapper($db, $request);
         $newClient = $mapper->getClientByIdAndUser($clientId, $request->user_id);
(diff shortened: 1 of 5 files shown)


iansltx:
How about instead making ApiView setters fluent? That would be a step in the direction of compatibility with PSR-7 (the step after would be adding immutability and changing set to with), as from what I can tell responses are treated as immutable, and ApiView is behaving rather like a response object here.

localheinz:
@iansltx 

See https://ocramius.github.io/blog/fluent-interfaces-are-evil/.

localheinz:
@iansltx 

In light of #783 I think it would be better to

- untangle the current `Request` object
- use a PSR-7 implementation (perhaps [`nyholm/psr-7`](https://github.com/Nyholm/psr7) or [`zendframework/zend-diactoros`](https://github.com/zendframework/zend-diactoros), although the latter probably doesn't make too much sense in the context of the shift from Zend Framework to Laminas)

What do you think?

iansltx:
Not sure why Diactoros would get DQ'd; mwop is going back to work for ~RogueZendForce~ Perforce on this sort of thing.

I personally prefer the Slim PSR-7 impl, but at the endo f the day it doesn't matter, as we'd be typehinting against either ServerRequestInterface or a decorated version thereof anyway (e.g. via slim/http).

iansltx:
To be clear (since I wasn't above at all), I'm in full support of getting the Request object to do Request-y things only, and using a PSR-7 impl for it. That does mean figuring out how we're going to negotiate content-types, but I think we could look at some middleware-ish pattern to do that job.

svpernova09:
I feel like there's bigger / better things to be done? Seems pendantic.

PR #799, opened 2019-11-11
Fix: Remove useless else statements

localheinz:
This PR

* [x] removes useless `else` statements


localheinz:
Thank you, @svpernova09!

PR #800, opened 2019-11-11

localheinz:
This PR

Follows #726.


localheinz:
Thank you, @svpernova09!

PR #801, opened 2019-11-11
Enhancement: Closures can be static

localheinz:
This PR

* [x] declares `closures` which can be `static` as `static`


localheinz:
Thank you, @svpernova09!

PR #802, opened 2019-11-11
Fix: Declare visibility on constants

localheinz:
This PR

* [x] declares visibility on constants


localheinz:
Thank you, @svpernova09!

PR #803, opened 2019-11-11
Enhancement: Use named constants instead of magic strings for HTTP methods

localheinz:
This PR

* [x] uses named constants instead of magic strings for HTTP methods
* [x] uses consistent wrapping


localheinz:
Thank you, @svpernova09!

PR #806, opened 2019-11-12
Fix: Remove useless elseif statements

localheinz:
This PR

* [x] removes useless `elseif` statements


localheinz:
Thank you, @svpernova09!

PR #807, opened 2019-11-12
Enhancement: Raise phpstan level from 4 to max

localheinz:
This PR

* [x] raises the `phpstan` level from `4` to `max`
* [x] regenerates the baseline

Replaces #773.


localheinz:
Thank you, @svpernova09!

PR #808, opened 2019-11-13
Enhancement: Expose happy path

localheinz:
This PR

* [x] inverts conditions and throws or returns early to expose the happy path


localheinz:
Thank you, @svpernova09!

PR #809, opened 2019-11-13
Fix: Remove useless break statements

localheinz:
This PR

* [x] removes useless `break` statements


localheinz:
Thank you, @svpernova09!

PR #811, opened 2019-11-16
PhpStan should ignore undefined Prophecy methods

svpernova09:


phpstan.neon
```diff
@@ -5,6 +5,7 @@ parameters:
 
     ignoreErrors:
         - '#Joindin\\Api\\Test\\Mock\\mockPDO::__construct\(\) does not call parent constructor from PDO\.#'
+        - '#Call to an undefined method Prophecy\\Prophecy\\ObjectProphecy::request\(\)#'
```
localheinz: There's something better that we could use once this turns out to be a problem (which it doesn't yet on `master`): [`jangregor/phpstan-prophecy`](https://github.com/Jan0707/phpstan-prophecy).

localheinz:
@svpernova09 

This can probably be closed in favour of #813.

PR #812, opened 2019-11-16
Enhancement: Generate and include baseline for phpstan

localheinz:
This PR

* [x] generates and includes a baseline for `phpstan`

❗️ Blocks #807.

💁‍♂ Run

```
$ composer stan-baseline
```

to regenerate `phpstan-baseline.neon`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,6 +73,11 @@
     "require-checker": "composer-require-checker --config-file=composer-require-checker.json",
     "security": "security-checker security:check composer.lock",
     "stan": "phpstan analyze --configuration=phpstan.neon",
+    "stan-baseline": [
+      "sed -e '/phpstan-baseline\\.neon/ s/^/#/' phpstan.neon > phpstan-without-baseline.neon",
+      "phpstan analyze --configuration phpstan-without-baseline.neon --error-format baselineNeon > phpstan-baseline.neon  || true",
+      "rm phpstan-without-baseline.neon"
+    ],
     "test": "phpunit"
   },
   "support": {
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,10 @@
+
+
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Joindin\\\\Api\\\\Test\\\\Mock\\\\mockPDO\\:\\:__construct\\(\\) does not call parent constructor from PDO\\.$#"
+			count: 1
+			path: tests/Mock/mockPDO.php
+
+
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,10 +1,10 @@
+includes:
+    - phpstan-baseline.neon
+
 parameters:
     excludes_analyse:
         - public/
         - scripts/
-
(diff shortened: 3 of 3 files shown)


localheinz:
Build failure is probably related to https://github.com/phpstan/phpstan/issues/2558.

ericpoe:
I did not know this was a feature of phpstan. I can see from https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff what it provides, and that looks great for all new code moving forward.

I have several questions:

- How do we know to bring existing code up to spec so we can reduce the number of exceptions-to-the-rule in this new file?
- While refactoring something that is excepted in this baseline, what is the best way to tell PHPStan that now is the time when it's OK to flag something that is in the baseline file?
- What are the best practices that you have found when working with this?

localheinz:
@ericpoe 

>How do we know to bring existing code up to spec so we can reduce the number of exceptions-to-the-rule in this new file?

We can tackle errors one by one.

Generally speaking, most errors can be avoided by 

- writing strict, defensive code
- making use of the type system (and adding new types)
- removing useless conditions and statements

>While refactoring something that is excepted in this baseline, what is the best way to tell PHPStan that now is the time when it's OK to flag something that is in the baseline file?

 `phpstan` will terminate with a non-zero exit code when we 

- fix an error that is ignored via  `phpstan-baseline.neon`
- introduce an error that is not ignored via `phpstan-baseline.neon`

The baseline can be regenerated anytime by running

```
$ composer stan-baseline
```

However, it would be in our best interest to let `phpstan-baseline.neon` shrink over time.

>What are the best practices that you have found when working with this?

In personal projects, I do not use the baseline feature.

Instead, I use the `max` level by default, along with additional, stricter rules, for example

- [`phpstan/phpstan-deprecation-rules`](https://github.com/phpstan/phpstan-deprecation-rules)
- [`phpstan/phpstan-strict-rules`](https://github.com/phpstan/phpstan-strict-rules)
- [`localheinz/phpstan-rules`](https://github.com/localheinz/phpstan-rules)

When I encounter a few errors that I believe should be ignored for now (perhaps to avoid BC breaks), I list these in `phpstan.neon`.

In work projects, most of which involve legacy code bases, I use this feature. There, I like to fix issues one by one, or kinds of issues one by one.

Does this answer your questions?

localheinz:
@ericpoe 

I have adjusted #807 and now raise the `phpstan` level to `max`.

When you look at the diff you get an idea of the things that need to be done.

localheinz:
@ericpoe 

I have to correct myself: I just started using the baseline feature for `phpstan` in https://github.com/localheinz/github-changelog/pull/318.

ericpoe:
@localheinz This looks great! I did some experimentation with using the baseline and I think this will be a good move.

I'll create a ticket to add instructions on how and when to rebuild the PHPStan baseline file (when: when a refactor fixes a problem contained within)

localheinz:
Thank you, @ericpoe!

PR #813, opened 2019-11-16
Enhancement: Require jangregor/phpstan-prophecy

localheinz:
This PR

* [x] requires `jangregor/phpstan-prophecy`

Replaces #811.


localheinz:
Thank you, @mcneely!

PR #814, opened 2019-11-16
Better Error Logging

mcneely:
If request is empty the original exception never gets logged. This assumes the issue happened before or during request instantiation.


mcneely:
> Looks like you could do this same change with fewer revised lines by wrapping L22-L28 in an instanceof check for Request (which is better than a null check here anyway).

I update it slight to add instanceof but left it in pretty much the same order.

PR #815, opened 2019-11-17
Fix: Remove unused composer.phar

localheinz:
This PR

* [x] removes an apparently unused `composer.phar`


localheinz:
Thank you, @svpernova09!

PR #785, opened 2019-11-05
Enhancement: Use friendsofphp/php-cs-fixer instead of squizlabs/php_codesniffer

localheinz:
This PR

* [x] removes `squizlabs/php_codesniffer`
* [x] removes `jakub-onderka/php-parallel-lint`
* [x] requires `friendsofphp/php-cs-fixer`
* [x] adds a basic configuration for `php-cs-fixer`
* [x] adds a `composer` script for running `php-cs-fixer`

Related to #784.

PR #816, opened 2019-11-19
Enhancement: Enable and configure fixers without effect on code base

localheinz:
This PR

* [x] enables and configures all fixers which do not have an effect on the code base

Follows #785.

💁‍♂ This provides a good baseline from which we can move forward and enable and configure more fixers.


svpernova09:
Merging because the only failure was CodeCov, which I don't really care about.

localheinz:
Thank you, @svpernova09!

PR #818, opened 2019-11-21
Fix: Actually run php-cs-fixer in build

localheinz:
This PR

* [x] actually runs `friendsofphp/php-cs-fixer` on GitHub actions

Follows #785.

🤦‍♂ Sorry, my bad.


localheinz:
Ha, forgot that we are using StyleCI.

PR #819, opened 2019-11-21
Enhancement: Update CONTRIBUTING.md with instructions

### LINKED ISSUE #817: Update PHPStan instructions to use PHPStan Baseline file
Now that #812 is merged in, we need to add instructions for how and when to update the PHPStan baseline file.

When:
1. After refactoring an error found in the baseline file, a PHPStan scan will return a non-zero error along with a message about "Ignored error pattern ... was not matched in reported errors."
1. Fix any other errors listed
1. Then recreate the baseline file via `composer stan-baseline`
1. Commit the change and push as part of the PR.

localheinz:
This PR

* [x] updates `CONTRIBUTING.md` with instructions for using `composer` scripts

Fixes #817.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,3 +45,69 @@ is not used as a work-in-progress list.
 When you have time to address the comments left on your PR,
 please make the changes, push them to github and then re-open
 the pull request.
+
+## GitHub Actions
+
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @svpernova09!

PR #820, opened 2019-11-21
Fix: Drop support for PHP 7.2

localheinz:
This PR

* [x] drops support for PHP 7.2

💁‍♂ Are we using PHP 7.3 in production? Then we probably don't need to be able to run on PHP 7.2.


localheinz:
Thank you, @svpernova09!

PR #823, opened 2019-11-25
Fix: Simplify baseline regeneration

localheinz:
This PR

* [x] simplifies the regeneration of the baseline for `phpstan`


localheinz:
Thank you, @svpernova09!

PR #827, opened 2019-12-03
Enhancement: Enable and configure binary_operator_spaces fixer

localheinz:
This PR

* [x] enables and configures the `binary_operator_spaces` fixer
* [x] runs `composer cs`

💁‍♂ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.16.1#usage:

>**binary_operator_spaces** [`@Symfony`, `@PhpCsFixer`]
>
>Binary operators should be surrounded by space as configured.
>
>Configuration options:
>
>* `align_double_arrow` (`false`, `null`, `true`): whether to apply, remove or ignore double arrows alignment; defaults to `false`. DEPRECATED: use options `operators` and `default` instead
>* `align_equals` (`false`, `null`, `true`): whether to apply, remove or ignore equals alignment; defaults to `false`. DEPRECATED: use options `operators` and `default` instead
>* `default` (`'align'`, `'align_single_space'`, `'align_single_space_minimal'`, `'no_space'`, `'single_space'`, `null`): default fix strategy; defaults to `'single_space'`
>* `operators` (`array`): dictionary of binary operator => fix strategy values that differ from the default strategy; defaults to `[]`

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -23,6 +23,9 @@ return PhpCsFixer\Config::create()
             'syntax' => 'short',
         ],
         'backtick_to_shell_exec' => true,
+        'binary_operator_spaces' => [
+            'default' => 'single_space',
+        ],
         'blank_line_after_opening_tag' => true,
         'blank_line_before_statement' => [
             'statements' => [
--- a/inc/nginx-helper.php
+++ b/inc/nginx-helper.php
@@ -10,12 +10,12 @@
  */
 function apache_request_headers()
 {
-    $arh     = [];
+    $arh = [];
     $rx_http = '/\AHTTP_/';
 
     foreach ($_SERVER as $key => $val) {
         if (preg_match($rx_http, $key)) {
-            $arh_key    = preg_replace($rx_http, '', $key);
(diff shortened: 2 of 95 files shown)


localheinz:
@svpernova09 

Something is entirely wrong here with the StyleCI integration - as pointed out before, the build should have failed here, but instead it passes. 

Looking at https://docs.styleci.io/configuration the following would need to be done:

- provide a `.styleci.yml` configuration file (which effectively duplicates the configuration in `.php_cs`) **or**
- manually configure the fixers on the StyleCI web interface

Personally, I would not want to manage two configuration files when one suffices. 

I remember seeing some kind of bridge that allowed using `.styleci.yml` to run `php-cs-fixer` locally, but I don't really think that this makes a lot of sense, as it's just another layer of obfuscation.

Looking at https://packagist.org/packages/styleci/ I see that most packages are abandoned. Perhaps it's time to revisit #818 and say good bye to StyleCI?

localheinz:
@heiglandreas @svpernova09 

Can you take a look?

PR #920, opened 2022-08-31
Tweaks for adding hosts to events via UI

heiglandreas:
This commit contains some minor tweaks to make handling multiple hosts via the UI easier.

It mainly disallows adding one host multiple times.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,11 @@
       "php": "7.3.11"
     },
     "preferred-install": "dist",
-    "sort-packages": true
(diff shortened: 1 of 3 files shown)


iansltx:
Once StyleCI passes, we're good to merge here, though we should also drop the Symfony security scanner as it apepars to be no more?

heiglandreas:
I'd leave dropping the Symfony security scanner (replacing with the Roave one?) to a separate PR. Along with replacing travis with GithubActions...

PR #945, opened 2023-05-26
Evaluate more-specific talk comment report/moderation routes ahead of less-specific talk comment update routes

iansltx:
This fixes talk comment report moderation being shadowed by talk comment updates due to an overly-wide regex for the latter. One could argue that the proper way to fix this is by fixing the regex for the latter, but I'm not comfortable enough with regex-fu to attempt that, and since our router evaluates routes one by one, in order, this will do the trick.

Technically we don't need to move *both* report routes, as POST has no routes shadowing it, but having two talk comment routes separated in the code looks weird enough that someone might attempt to reorder the routes back, causing the bug to resurface.

This bug manifests itself as a 500 on web2 when attempting to take action on a talk comment report, wrapping a 400 on the API side, with the API side complaining about "comment" not being supplied when required (which is a requirement of the comment update action but not the comment moderation action).

### DIFF
--- a/src/config/routes/2.1.php
+++ b/src/config/routes/2.1.php
@@ -454,22 +454,6 @@ return [
         ],
     ],
     [
-        'path'       => '/talk_comments(/[^/]+)*/?$',
-        'controller' => TalkCommentsController::class,
-        'action'     => 'getComments',
-        'verbs'      => [
-            Request::HTTP_GET,
-        ],
-    ],
-    [
-        'path'       => '/talk_comments(/[^/]+)*/?$',
-        'controller' => TalkCommentsController::class,
-        'action'     => 'updateComment',
-        'verbs'      => [
-            Request::HTTP_PUT,
-        ],
-    ],
-    [
         'path'       => '/talk_comments(/[^/]+)*/reported?$',
         'controller' => TalkCommentsController::class,
         'action'     => 'reportComment',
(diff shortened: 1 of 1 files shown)


iansltx:
The diff is janky here. What _actually_ happened is I moved

```php
    [
        'path'       => '/talk_comments(/[^/]+)*/reported?$',
        'controller' => TalkCommentsController::class,
        'action'     => 'reportComment',
        'verbs'      => [
            Request::HTTP_POST,
        ],
    ],
    [
        'path'       => '/talk_comments(/[^/]+)*/reported?$',
        'controller' => TalkCommentsController::class,
        'action'     => 'moderateReportedComment',
        'verbs'      => [
            Request::HTTP_PUT,
        ],
    ],
```

from immediately below

```php
    [
        'path'       => '/talk_comments(/[^/]+)*/?$',
        'controller' => TalkCommentsController::class,
        'action'     => 'getComments',
        'verbs'      => [
            Request::HTTP_GET,
        ],
    ],
    [
        'path'       => '/talk_comments(/[^/]+)*/?$',
        'controller' => TalkCommentsController::class,
        'action'     => 'updateComment',
        'verbs'      => [
            Request::HTTP_PUT,
        ],
    ],
```

to immediately above that block

iansltx:
Fun part of this is, _I_ tripped up on positioning these routes when I added them back in https://github.com/joindin/joindin-api/commit/b40605dae4b9132ac07f848fe0ff34b05e0c640c, as before they were removed in https://github.com/joindin/joindin-api/commit/fbf69cc32fd3382108b2c0feebe42749af4c1b4c they were split, such that the PUT route was in the correct place to work, and the placement of the POST route didn't matter.

svpernova09 (approved):

Issue #916, opened 2021-06-22
No way to change username


OGProgrammer:
I have an old an undesirable username, however I see no way to change it via the web app.

Could I create a migration for my specific username or build the feature to actually change usernames?

I login with twitter but I've since changed my twitter username but didn't pick up here.

Thanks!

iansltx:
It's been awhile, but I think there were some concerns around user-editable usernames.

With that said, since doing so would provide an audit trail, feel free to comment here with the username change migration. Don't think we want it in the codebase per se as in theory you could run joind.in from a blank database elsewhere, but that doesn't stop us from running the statements manually.

heiglandreas:
Hey.

That change a username has so far been cause for discussions every now and then. So far we solved that usually by editing the database directly in the rare cases where that was requested. 

But having that as a "self-service feature" (and if only via the API in the first place) would be interesting. Our discussions usually stopped at the point where we weren't sure what to do with the "old" username. To prevent that name from being taken by someone else and used possibly to the disadvantage of the former "owner" that name should definitely be reserved. And possibly (in the case of a "propper" implementation) also be redirected to the new name. At least if the former owner wishes to do so.

So should I decide to change my username from `heiglandreas` to `heigla` any request to `/users/heiglandreas` should do a redirect to `/users/heigla` if I requested such a redirect.

So it would require a lookup-possibility for `old username` to know the `new username` which also could contain a flag `redirect`.

And all usernames from the `old username` would need to be considered when checking for an already registered username.

The main concern was that the feature could be abused to block usernames without creating multiple users by simply changing usernames multiple times. And that was usually the point where the discussion stopped.

Technically it's solvable, but – as usual – the challenge comes with the criminal energy some folks invest in breaking cool things.  :frowning:

heiglandreas:
See #357

OGProgrammer:
My old username is `PsycodeDotOrg` and want it updated to `OGProgrammer` - Even if it doesn't redirect, it is fine by me. I use twitter to login and that was my old twitter username which has since been updated.

iansltx:
Given @heiglandreas's note, seems like you could grab an API key and make the change with the codebase that's currently deployed, without needing to make upgrades otherwise. Mind trying this and reporting back? If it works, that should at least get you un-stuck.

OGProgrammer:
[...]

PR #947, opened 2023-07-19
Allow renaming username

heiglandreas:
This checks that there actually is a user returned with the new name and if not just allow the renaming.

THis fixes the issue described in https://github.com/joindin/joindin-api/issues/916#issuecomment-1642395390


zghosts (approved):
Looks ok to me

nicwortel:
@iansltx Is there any plan to release a new version containing this fix? Would like to be finally able to change my username :innocent:

